Issue #10, opened 2019-12-16
UCB1 arm_to_expectation not updated for all arms (potential bug)


harisankarh:
When `partial_fit()` is done with UCB1, `arm_to_expectation` of only the arms having a reward are updated (because of [https://github.com/fmr-llc/mabwiser/blob/0c860253be017d1f393e18bf9d9d7e1739f93dca/mabwiser/ucb.py#L62](https://github.com/fmr-llc/mabwiser/blob/0c860253be017d1f393e18bf9d9d7e1739f93dca/mabwiser/ucb.py#L62) ). If an arm does not have a reward, its `arm_to_expectation` is not updated.

`arm_to_expectation` depends on `self.total_count` which gets updated when "any" of the arms are invoked. Thus, `arm_to_expectation` of all arms need to be updated when `self.total_count` changes.

Solution: remove the above condition (`if arm_rewards.size`).

Happy to submit a fix if this change can be made.

erstrong:
This is not a bug. In the expectation calculation for UCB1 you divide by the number of observations of the arm (see [documentation](https://fmr-llc.github.io/mabwiser/api.html#mabwiser.mab.LearningPolicy.UCB1) for the equation). For a new arm that number is zero, so we cannot divide by it and must leave the expectation as zero.

From a conceptual standpoint, you can think of it as if we don't have any data for an arm we cannot calculate an upper confidence bound on its unknown mean.

skadio:
@harisankarh thank you for bringing this to our attention! As @erstrong mentioned, hope the UCB1 formula clarifies this issue. 

Thank you again for your interest in our library!

harisankarh:
@erstrong @skadio Thanks for the response

@erstrong  This issue can arise even in cases other than "new arm or arm with no data" which you mentioned. 
Consider a bandit with arms `a`, `b` and `c`. 
First, `a` is invoked 10 times and the total reward obtained for `a` is 5. `a`'s `arm_to_expectation` is updated accordingly. 
[...]

PR #12, opened 2019-12-17
Bug fix: update expectations in ucb1 partial fit when no new data for arm

erstrong:
Addresses bug reported in #10 

The _fit_arm function now checks arm_to_count[arm] instead of size to update the expectation for the arm. This change avoids a divide by zero error but ensures that arms without new data in a partial fit will still have their expectations updated based on the new total number of observations. Unit test for partial fit method also updated to check an arm for which this case occurs.


CHANGELOG.txt
```diff
@@ -2,6 +2,13 @@
 MABWiser CHANGELOG
 =====================
 
+
+-------------------------------------------------------------------------------
+December, 17, 2019 1.7.1
+-------------------------------------------------------------------------------
+minor:
+- Bug fix for partial fitting in UCB1
```
skadio: in here, can we acknowledge/thank @xx for bringing this bug to our attention

erstrong: Sure!

PR #14, opened 2020-01-16
LinTS

erstrong:
Adds LinTS algorithm and refactors random generator usage. Adds version dependency for scikit-learn due to a bug fix for KMeans in 0.22.


mabwiser/mab.py
```diff
@@ -67,6 +67,68 @@ def _validate(self):
...
+            'Arm2'
+        """
+        alpha: Num = 1.0
+        l2_lambda: Num = 1.0
+        arm_to_scaler: Dict[Arm, Callable] = None
+
+        def _validate(self):
+            check_true(isinstance(self.alpha, (int, float)), TypeError("Alpha must be an integer or float."))
+            check_true(0 < self.alpha, ValueError("The value of alpha must be greater than zero."))
+            check_true(isinstance(self.l2_lambda, (int, float)), TypeError("L2_norm must be an integer or float."))
```
bkleyn: Typo here. Change "L2_norm" to "l2_lambda".

mabwiser/mab.py
```diff
@@ -565,7 +632,8 @@ def __init__(self,
         self.backend = backend
 
         # Create the random number generator
-        self._rng = np.random.RandomState(seed=self.seed)
+        # self._rng = np.random.RandomState(seed=self.seed)
```
bkleyn: Can probably remove this commented line.

mabwiser/linear.py
```diff
@@ -45,33 +45,62 @@ def fit(self, X, y):
 
         # Update A
         self.A = self.A + np.dot(Xt, X)
-        self.A_inv = np.linalg.inv(self.A)
+
+        try:
+            self.A_inv = np.linalg.inv(self.A)
+        except np.linalg.LinAlgError:
```
bkleyn: Does it make sense to return coefficients if A is not invertible?

mabwiser/base_mab.py
```diff
@@ -165,7 +165,7 @@ def _parallel_predict(self, contexts: np.ndarray, is_predict: bool):
         total_contexts = sum(n_contexts)
 
         # Get seed value for each context
-        seeds = self.rng.randint(np.iinfo(np.int32).max, size=total_contexts)
+        seeds = RandomGenerator.get_rng_seeds(self.rng, total_contexts)
```
bkleyn: Nice!

mabwiser/linear.py
erstrong: Reverted to old version

examples/parametric_mab.py
```diff
@@ -109,3 +109,74 @@
...
 print("KNearest: ", prediction, " ", expectations)
 assert(prediction == [1, 2])
+
+##################################################
+# Linear Thompson Sampling Learning Policy
+##################################################
+
+# LinTS learning policy with alpha 1.25 and l2_lambda 1
+lints = MAB(arms=ads,
+             learning_policy=LearningPolicy.LinTS(alpha=1.5, l2_lambda=1))
```
skadio: extra whitespace

examples/parametric_mab.py
```diff
@@ -109,3 +109,74 @@
 # Results
 print("KNearest: ", prediction, " ", expectations)
 assert(prediction == [1, 2])
+
```
skadio: these usage examples are nice --thanks!

mabwiser/linear.py
```diff
@@ -45,6 +45,7 @@ def fit(self, X, y):
 
         # Update A
         self.A = self.A + np.dot(Xt, X)
+
```
skadio: is this intentional?

mabwiser/linear.py
```diff
@@ -57,21 +58,51 @@ def predict(self, x):
...
         # Scale
         if self.scaler is not None:
-            x = x.reshape(1, -1)
-            x = self.scaler.transform(x.astype('float64')).reshape(-1)
+            x = self._scale_predict_context(x)
 
         # Calculate default expectation y = x * b
         return np.dot(x, self.beta)
 
+    def _scale_predict_context(self, x):
```
skadio: can we add comments to this helper functions.. what does reshape(1, -1) does, transforms? floats?

mabwiser/linear.py
```diff
@@ -57,21 +58,51 @@ def predict(self, x):
...
+
+class _LinTS(_RidgeRegression):
+
+    def predict(self, x):
+
+        # Scale
+        if self.scaler is not None:
+            x = self._scale_predict_context(x)
+
+        # Randomly sample coefficients from Normal Distribution N(mean=beta, variance=exploration)
```
skadio: does this comment match the lin?

mabwiser/linear.py
```diff
@@ -57,21 +58,51 @@ def predict(self, x):
...
+        exploration = np.square(self.alpha) * self.A_inv
+
+        beta_sampled = self._multivariate_normal(exploration)
```
skadio: May be a quick one-liner comment here?

mabwiser/linear.py
```diff
@@ -57,21 +58,51 @@ def predict(self, x):
...
+
+        # Calculate expectation y = x * beta_sampled
+        return np.dot(x, beta_sampled)
+
+    def _multivariate_normal(self, exploration):
+        # Multivariate Normal Sampling
```
skadio: can you add a quick word on what's exploration parameter? int float constant? array/matrix?

mabwiser/mab.py
```diff
@@ -67,6 +67,68 @@ def _validate(self):
             check_true(isinstance(self.epsilon, (int, float)), TypeError("Epsilon must be an integer or float."))
             check_true(0 <= self.epsilon <= 1, ValueError("The value of epsilon must be between 0 and 1."))
 
+    class LinTS(NamedTuple):
```
skadio: Very nice docstring!

mabwiser/linear.py
```diff
@@ -162,10 +193,14 @@ def _predict_contexts(self, contexts: np.ndarray, is_predict: bool,
         # Create an empty list of predictions
         predictions = [None] * len(contexts)
         for index, row in enumerate(contexts):
+            rng = np.random.RandomState(seed=seeds[index])
```
skadio: This line needs to be commented well. 

Spelling out the control flow for learning purposes.. So this class itself has a "self.rng"... but we don't use that here.. and instead, look at the "seeds" which is given as input.

Then, we create a *new* rng based on the particular seed... we then copy that to all arms.. 

Previous _Linears, linUCB and Ridge did not need that BUT LinTS needs it?

Should these only happen for LinTS? 

CC @bkleyn

mabwiser/linear.py
erstrong: It's referring to the next two lines

mabwiser/linear.py
erstrong: Ridge and LinUCB don't have any random behavior

examples/parametric_mab.py
erstrong: What is this comment referring to?

mabwiser/linear.py
bkleyn: Previously _RidgeRegression and _LinUCB had their own rng objects, but it wasn't used for anything. (I'm not exactly sure why this was the case given that these classes do not inherit from BaseMAB?) But yes, now we create a new rng for each context using seeds that were generated from the "master" rng held by BaseMAB. 

I don't think the fact that the _RidgeRegression and _LinUCB classes do not use the rng object should be an issue. The same is true for the _UCB1 class that has a rng object that's not used in the learning policy.

erstrong: _RidgeRegression and _LinUCB have rng objects because _LinTS was originally intended to be part of our initial release. When we removed it we removed only that class and left _RidgeRegression unchanged. Since neither _RidgeRegression nor _LinUCB uses the rng object, it was left out of the changes made to add parallelization. However _LinTS does use the rng, so we now need to add this change which mirrors how we handle the parallel rngs in the _NearestNeighbors classes.

skadio: Thanks for the clarification!

PR #16, opened 2020-05-02
Popularity

skadio:
Randomized Popularity Bandit Implementation + Tests 

At a high-level:
- It utilizes the _Greedy bandit under the hood, exactly the same
- The difference is, it normalizes the arm_to_expectation to be probabilities that sum up 1.0 
- Rng.Choice() is used to predict() a randomized arm based on popularity weights/probabilities


mabwiser/mab.py
```diff
@@ -189,10 +190,32 @@ def _validate(self):
             if self.arm_to_scaler is not None:
                 check_true(isinstance(self.arm_to_scaler, dict), TypeError("Arm_to_scaler must be a dictionary"))
 
+    class Popularity(NamedTuple):
+        """Randomized Popularity Learning Policy.
+
+        Returns a randomized popular arm for each prediction.
```
erstrong, line 196: This should include the formula for calculating the distribution

mabwiser/mab.py
```diff
@@ -701,7 +727,10 @@ def learning_policy(self):
             lp = self._imp
 
         if isinstance(lp, _EpsilonGreedy):
-            return LearningPolicy.EpsilonGreedy(lp.epsilon)
+            if issubclass(type(lp), _Popularity):
```
erstrong, line 730: Nice catch, though you could simplify the logic by checking for Popularity as the first option instead of sticking to the alphabetical order (if Popularity, elif EpsilonGreedy)

tests/test_popularity.py
```diff
@@ -0,0 +1,195 @@
+# -*- coding: utf-8 -*-
+
+from mabwiser.mab import LearningPolicy, NeighborhoodPolicy
+from tests.test_base import BaseTest
+
+
+class PopularityTest(BaseTest):
```
erstrong, line 7: Since Popularity inherits from EpsilonGreedy it also inherits the epsilon property. Should there be a unit test to verify that manually changing epsilon has no impact on the predictions?

mabwiser/popularity.py
```diff
@@ -0,0 +1,48 @@
...
+        # Make sure expectations sum up to 1 like probabilities
+        self._normalize_expectations()
+
+    def predict(self, contexts: np.ndarray = None) -> Arm:
+
+        # Select an arm randomized by expectation probability
+        return self.rng.choice(self.arms, p=list(self.arm_to_expectation.values()))
+
+    def _normalize_expectations(self):
+        # TODO: this would not work for negative rewards!
```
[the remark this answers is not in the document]
skadio: no, I don't plan to address this as we don't have an immediate case for it, but it's a warning in case someone comes here. We shall address it though eventually.

mabwiser/popularity.py
```diff
@@ -0,0 +1,48 @@
...
+# SPDX-License-Identifer: Apache-2.0
+
+from typing import List, Optional, NoReturn
+import numpy as np
+
+from mabwiser.greedy import _EpsilonGreedy
+from mabwiser.utils import Arm, reset
+
+
+class _Popularity(_EpsilonGreedy):
```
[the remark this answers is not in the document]
skadio: Nothing in particular, other than most bandits sits on their own including random

mabwiser/mab.py
skadio: I didn't see one for Greedy so I skipped it, but that's a good point

mabwiser/mab.py
skadio: I like that --let me assert that it works as you intended, and I will revise

tests/test_popularity.py
skadio: Good point! I will add

mabwiser/popularity.py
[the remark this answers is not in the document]
skadio: Btw, I don't know what that algorithm for the case of negative rewards should look like. It's not obvious to me.. Maybe shift everything by min(rewards) first, and scale later.. Or see the L1 or L2 versions.. Each will probably have different behaviour/performanece and it's not straightforward to me how that should it look like. Something to think in the background

mabwiser/mab.py
skadio: I checked that it works as you suggested with Popularity as the first option: 

        if isinstance(lp, _Popularity):
            return LearningPolicy.Popularity()
        elif isinstance(lp, _EpsilonGreedy):
            return LearningPolicy.EpsilonGreedy(lp.epsilon)

My only issue with this style is, if someone reorders this if-else block, then tests will fail. Capturing the block wiht if (_Greedy) catch makes it more explicit, and then implicit order above. 

Plus, that's consistent with how (_Linear) catches among different regressions.

Explicity is better than implicit. So I recommended keeping it as is

mabwiser/popularity.py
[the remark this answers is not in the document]
erstrong: Shifting by the min makes sense. Perhaps this TODO should be moved to be a line in the API documentation? Also, it might be good to add a unit test to verify what the behavior is when some of the arms have negative means

mabwiser/popularity.py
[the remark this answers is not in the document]
erstrong: The other context-free algorithms are separate because they don't inherit from each other, but the nearest neighbors ones and the linear ones are together.

tests/test_popularity.py
skadio: Btw, notice that standard bandits mab = MAB(LP.Popularity()) would allow you to access epsilon manually :) We close that door long time ago.
I can still create a hacky test .

mabwiser/mab.py
erstrong: For greedy it's just calculating the means, but here we are generating a distribution

mabwiser/popularity.py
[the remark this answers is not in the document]
skadio: I see, so is your recommendation to move it under _Greedy?

tests/test_popularity.py
erstrong: I think there are two issues - LP.Popularity does not have an epsilon value and there should be a test in test_invalid to verify that if you try to pass an epsilon value to LP.Popularity it raises an error, but mabwiser.popularity.Popularity does have epsilon that could be accessed as mab._imp.epsilon so we should verify that nothing happens if the person accesses the implementor to change that value

mabwiser/popularity.py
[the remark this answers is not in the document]
erstrong: I think that would make the most sense

tests/test_popularity.py
skadio: done, and done for both tests.

skadio:
Thanks for @erstrong for the detailed review comments! Addressed them below. 

Add algorithm to the supported lps for _NeighborsSimulator, _RadiusSimulator, _KNearestSimulator
>> Done. Btw, the existing order was weird/random --now fixed. 

Add algorithm to the context-free examples
>> Good point, done! 

Add unit tests to test_examples for the code added to examples
>> Good point, done! 

Add algorithm to the lps for the TestSimulator class in test_simulator
>> Done. Btw, why is test_simulator not using the LPs and NPs in test_base? and has its own separate version? To keep the lists, shorter? 

Possibly need to add unit tests to test_parallel
>> Good point, done!

The popularity unit tests are missing the following standard unit tests that all of the other algorithms have: fit twice, partial fit, unused arm, add arm, tests for different input formats, tests for different decision types (t1-t10)
>> I doubt all policies have this, see e.g. random, but in any case, isn't this what test_mab is checking for as it loops through LP list? 

Update version/release number in setup.py and docsrc/conf.py. This second one is how the docs pages get the release number.
>> ah thanks! docsrc/conf.py is news to me.. I didn't know we had a version there. Side note: we might want to create a script "update_version.sh" to bump up the counters under /scripts.. 

Update the changelog and recompile the wheel file
>> ah missed that, thanks, done!

skadio:
on a separate note 
@erstrong --you not only provided a detailed review but also just gave me a great idea! Most of these review comments can serve as a "template" if any wants to contribute to mabwiser. 

I knew some of these (like changing versions here and there and compiling a new wheel etc) but then, it's easy to miss these. 

I need your help to create a "template" review check list. Let me follow up on that separately.

mabwiser/popularity.py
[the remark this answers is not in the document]
skadio: I can do that, but my only concern would be this: 

if I am a user and I am curious about the implementation behind the popularity algorithm, which is basically us a month ago when we were trying to find a popularity implementation, it would be nice to come here, browse mabwiser folder and go directly to populairty.py.. 

If I am after the implementation of Neighborhood policies.. I can go to neighborhood.py
If I am after the implementation of LinTS/LinUCB policies.. I can go to linear.py

But the fact that I need to go to "greedy.py" for the implementation of popularity is not apparent anywhere.. Unless I do a some form of search or sth. So the connection between greedy and popularity is not naturally there for me. It is only an implementation connection for only those who knows how things are implemented under the hood. 

That's my reasoning to keep somewhat of a mapping between bandits and the files here. Popularity = popularity.py

Thoughts?

mabwiser/mab.py
skadio: done

mabwiser/popularity.py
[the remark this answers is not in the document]
erstrong: Sounds reasonable to me

erstrong:
> Done. Btw, why is test_simulator not using the LPs and NPs in test_base? and has its own separate version? To keep the lists, shorter?

The simulator tests don't iterate through the LPs the way we do in other unit tests, the tests run simulations with all of them as bandits

erstrong:
> The popularity unit tests are missing the following standard unit tests that all of the other algorithms have: fit twice, partial fit, unused arm, add arm, tests for different input formats, tests for different decision types (t1-t10)
> 
> > > I doubt all policies have this, see e.g. random, but in any case, isn't this what test_mab is checking for as it loops through LP list?

Random is the only one without them because it doesn't perform any actual calculations. test_mab doesn't check all of these, and it doesn't look at the exact values. These unit tests are how we catch changes that impact the outputs of the algorithms (e.g. the change to KMeans in the latest version of sklearn broke several test_clusters tests but nothing in test_mab).

skadio:
@erstrong 
Let me add more tests for your #6 above

> The popularity unit tests are missing the following standard unit tests that all of the other algorithms have: fit twice, partial fit, unused arm, add arm, tests for different input formats, tests for different decision types (t1-t10)

docs/_static/jquery.js
```diff
@@ -1,2 +1,4 @@
...
\ No newline at end of file
+/*! jQuery v3.2.1 | (c) JS Foundation and other contributors | jquery.org/license */
```
erstrong, line 1: You need to manually swap this with the v3.5.0 that was overwritten. This is unfortunately something we have to do until Sphinx updates their version dependency, unless we aren't concerned about the security warning

skadio: Good point, but let me come back to this after this merge --to fix security warning, IF it shows up. I am using the same setup for Seq2Pat and it does not give a warning --so this might just work

skadio: Gotch you, that make sense. Let me go through with this PR and we can fix the security warning separately.

erstrong:
@skadio I don't see the following in the latest commits:
1. Add algorithm to the supported lps for _NeighborsSimulator, _RadiusSimulator, _KNearestSimulator in the simulator module
2. Standard unit tests for different decision types (t1-t10). For an example of these see lines 193-320 in test_greedy
Also looks like the version needs to be incremented in simulator.py

skadio:
@erstrong 

Add algorithm to the supported lps for _NeighborsSimulator, _RadiusSimulator, _KNearestSimulator in the simulator module

>> I missed these, thanks. Added now. Btw, can we plz keep these Union[] lists in alphabetic order going forward. They all seem to be mixed. 

Standard unit tests for different decision types (t1-t10). For an example of these see lines 193-320 in test_greedy. 
>> I looked at this yesterday as well, still not sure what it really means. I added the case for string arms, numpy rewards/decisions, and data series. Can you plz be specific about what other decision types are you referring to? 

Also looks like the version needs to be incremented in simulator.py
>> Good catch. Done

erstrong:
> Standard unit tests for different decision types (t1-t10). For an example of these see lines 193-320 in test_greedy.
> 
> > > I looked at this yesterday as well, still not sure what it really means. I added the case for string arms, numpy rewards/decisions, and data series. Can you plz be specific about what other decision types are you referring to?

The tests have different combinations of int decisions, string decisions, object decisions, and binary rewards, positive continuous rewards, and rewards that include negative numbers. For the other algorithms they also test different values of hyper-parameters (eg different epsilon) but since Popularity doesn't have any that wouldn't be necessary. 

Also, the new unit test in test_parallel is failing. It looks like you have a different list of arms for when n=1 than for the other two scenarios.

skadio:
> > Standard unit tests for different decision types (t1-t10). For an example of these see lines 193-320 in test_greedy.
> > > > I looked at this yesterday as well, still not sure what it really means. I added the case for string arms, numpy rewards/decisions, and data series. Can you plz be specific about what other decision types are you referring to?
> 
> The tests have different combinations of int decisions, string decisions, object decisions, and binary rewards, positive continuous rewards, and rewards that include negative numbers. For the other algorithms they also test different values of hyper-parameters (eg different epsilon) but since Popularity doesn't have any that wouldn't be necessary.

>> One parameter we can test is seed. Added another test for that. 

> Also, the new unit test in test_parallel is failing. It looks like you have a different list of arms for when n=1 than for the other two scenarios.

>> Thanks, fixed the test!

skadio:
@erstrong 
Added the changes we talked about. A few more tests, fixed docstest, reviesed new_bandit.rst

erstrong:
Not a major issue, but the unit test for negative values should use an assertRaises instead of a try/catch block. Otherwise looks good!

PR #20, opened 2020-05-27
LinTS performance improvement

bkleyn:
Moved cholesky decomposition to fit method to speed up prediction


mabwiser/linear.py
```diff
@@ -72,32 +72,47 @@ def _scale_predict_context(self, x):
...
-            x = self._scale_predict_context(x)
+        # Set covariance to none
+        self.covar_decomposed = None
 
-        # Calculate the covariance matrix multiplied by alpha squred
+    def init(self, num_features):
+        super().init(num_features)
+
+        # Calculate exploration factor
         exploration = np.square(self.alpha) * self.A_inv
```
erstrong, line 86: I recommend putting lines 86-89 and 95-98 in a helper function to avoid code duplication

skadio: That, and in case we use another decomposition (unlikely but..)

bkleyn:
> All unit tests are passing for me. Out of curiosity, how much of a speedup do you get with this change?

10-15x speedup for ~500 dimensional context.

mabwiser/linear.py
bkleyn: Added _cholesky helper function as suggested.

PR #21, opened 2020-06-02
spring cleaning

skadio:
sign

Signed-off-by: Kadioglu, Serdar <serdar.kadioglu@fmr.com>


scripts/build.md
```diff
@@ -1,48 +0,0 @@
-# Building the Package
```
skadio, line 1: Removing this because: 
* The part about the Install can be found in the IO docs
* The part about sphinx is not interesting from a user point of view externally and captured in Conf internally

PR #25, opened 2020-06-22
Update RNG

skadio:
The high-level idea of changes: make MABwiser agnostic to the underlying random number generator. 

To do that: 

* Identified current RNG needs. We currently use random float/int, beta/std distribution, and choice
* create a base rng that sets the contract for these operations 
* implement an RNG that uses Numpy.rand as the underlying generator for these operations
* Add a factory method (create_rng) that returns the concrete implementor 
* Let Mabwiser (mab.py) to initialize library-wide self.rng using the factory method. 

All tests pass as before --so there is no behavior change. 

In the future, if we want to upgrade our RNG to another implementation then: 

* Implement the new RNG class following the base rng contract
* Let the RNG factory return the new implementor


mabwiser/base_mab.py
```diff
@@ -61,12 +61,12 @@ class BaseMAB(metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def __init__(self, rng: np.random.RandomState, arms: List[Arm], n_jobs: int, backend: str = None):
+    def __init__(self, rng: _NumpyRNG, arms: List[Arm], n_jobs: int, backend: str = None):
```
erstrong: I think this should be _BaseRNG to allow for future changes to the library

mabwiser/utils.py
```diff
@@ -60,3 +62,132 @@ def reset(dictionary: Dict, value) -> NoReturn:
     """
     dictionary.update({}.fromkeys(dictionary, value))
 
+
+class _BaseRNG(metaclass=abc.ABCMeta):
```
erstrong, line 66: I think you need to add a seed function and change BaseMAB to call that. For example for default_rng the seeding method is:

seedseq = np.random.SeedSequence(seed)
seeds = seedseq.spawn(total_contexts)

mabwiser/utils.py
```diff
@@ -60,3 +62,148 @@ def reset(dictionary: Dict, value) -> NoReturn:
...
+        return self.rng.choice(a=a, size=size, p=p)
+
+    def beta(self, num_success: int, num_failure: int):
+        return self.rng.beta(num_success, num_failure)
+
+    def standard_normal(self, size):
+        return self.rng.standard_normal(size)
+
+
+def _create_rng(seed: int) -> _BaseRNG:
```
erstrong, line 196: If the requirement is that a _BaseRNG object be passed to BaseMAB constructors, this should be public

mabwiser/utils.py
```diff
@@ -60,3 +62,148 @@ def reset(dictionary: Dict, value) -> NoReturn:
     """
     dictionary.update({}.fromkeys(dictionary, value))
 
+
+class _BaseRNG(metaclass=abc.ABCMeta):
```
bkleyn, line 66: This is missing a method to get the state of the generator. See get_state method used in _Clusters class.

skadio: can you point where it is needed/used (I wonder why no tests complained..)

mabwiser/utils.py
```diff
@@ -60,3 +62,148 @@ def reset(dictionary: Dict, value) -> NoReturn:
...
+                A floating-point array of shape ``size`` of drawn samples
+        """
+        pass
+
+
+class _NumpyRNG(_BaseRNG):
+
+    def __init__(self, seed):
+        super().__init__(seed)
+        self.rng = np.random.RandomState(self.seed)
```
bkleyn, line 178: Think we'll need to try this with the new Numpy [Random Generator](https://numpy.org/doc/1.18/reference/random/generator.html) class to see if it works.

mabwiser/base_mab.py
skadio: Good catch! It should be _BaseRNG

mabwiser/utils.py
bkleyn: See clusters.py file line 30 and 32

erstrong: It was used in the Clusters class for copying the kmeans objects. You changed it to using the seed directly. However it's a standard function  that we may have future uses for

mabwiser/utils.py
skadio: Agreed! I made it public first, but could not find any usage so downgraded to private. 

You mentioned customized bandits, which would need this functionality! So I will make it public and update the examples for customized bandits.

mabwiser/utils.py
skadio: Previously, this used "rng.get_state()[1][0] " which is an extremely hacky way to just get the seed. 

Currently, we pull that as rng.seed which does the job (that's why no test failed) and seems clean to me. 

What is the suggestion exactly?

skadio: (Notice, we didn't need get_state() in any place, just the seed value..)

mabwiser/utils.py
skadio: Agreed. 

But may I define the scope of this PR narrower than that: replicate the current functionality as-is. Uptaking a new rng sounds like next update to me. 

That said if you two claim that this PR would not add much without also upgrading the RNG I can pull back. 

Thoughts @erstrong @bkleyn

mabwiser/utils.py
skadio: Agree that it's a desirable method, one we might need eventually. 

However, currently such functionality is not needed/used anywhere. So I am hesitant to add new features without know proper use cases.

mabwiser/utils.py
erstrong: We don't currently explicitly need it (agreed that the previous usage was hacky), but it's a common API function for random generators so it would be a good idea to have the function available

skadio: I completely agree and we will most likely that functionality very soon. I just don't want to add a functionality that is not going to be used now --unless you deem it as a "blocker" for this PR.

bkleyn: As long as we can directly access the seed from the random generator we don't currently have a need to get the state as I suggested.

mabwiser/utils.py
erstrong: My understanding is that the goal for this was to create a wrapper class that would allow us to swap out random generators without needing to change other code. The best way to do that is to test it with the Generator class, since there is at least one major difference that I am aware of (see my comment about seeds) that would require a function in the wrapper class, and there may be others.

mabwiser/utils.py
erstrong: Here is what I am referring to for the Generator (default_rng) class: https://numpy.org/devdocs/reference/random/parallel.html#seedsequence-spawn

bkleyn: Looking at this again, I'm not sure if we actually need to use these parallel seeding methods. Might be easiest to discuss offline.

mabwiser/utils.py
bkleyn: My concern was the uptaking the rng would not be trivial and that additional functionality would be needed to deal random states for parallelization. However, after a closer look I think it's simpler than we previously thought. So I'm good with merging this PR and then uptaking the new rng as a separate PR.

skadio:
@erstrong 

> To support custom bandits, I think the create_rng function should be public and these changes need to be added to examples/customized_mab.py

On this note. I take a closer look at customized_mabs. It is already making use of internal functions such as _Greedy and _Linear.. In that same behavior, it can also use _BaseRNG and _create_rng()

Thoughts?

mabwiser/utils.py
skadio: See my comment about this below. On a second look, I don't see why it should be public now. The examples definitely need to be updated to point out how to use _BaseRNG but just like how it uses _Greedy, _Linear it can use private functionality. I doubt we want this in the public API, since it is not used. 

Thoughts?

skadio:
Added a merge for the fixes in the reviews. 

- I didn't introduce any new functionality that wasn't needed in the library in its current state
- Updated customized bandit examples, but kept the _BaseRNG and _create_rng private similar to how customized bandits use _Greedy, _Linear etc.?

skadio:
@bkleyn thanks for the approval! Glad to hear we can uptake new generators. 

The bigger question would be 
* do we want that? 
* when do we want that? 
* what does that break from outside? 
* what does that break inside (tests etc.) 

I have a feeling that's the point when we will break from version 1.x.x to 2.x.x

bkleyn:
Couple of comments to your questions:

- From what I understand, the new [numpy generator](https://numpy.org/doc/1.18/reference/random/new-or-different.html#new-or-different) is 2-10 faster for generating random variables from the normal distribution (and distributions related to normal). This might speed up TS prediction. 
- If the legacy generator is no longer supported in numpy we'll have to provide support for the new generator in my view (or consider using the standard Python random module)
- Changing the generator would mean previous results (outside) will not be reproducible.
- Changing the generator will likely break most of the tests (inside).

Agreed that it will probably make sense to move to 2.x.x at this point.

erstrong:
> @bkleyn thanks for the approval! Glad to hear we can uptake new generators.
> 
> The bigger question would be
> 
> * do we want that?
> * when do we want that?
> * what does that break from outside?
> * what does that break inside (tests etc.)
> 
> I have a feeling that's the point when we will break from version 1.x.x to 2.x.x

@skadio I mentioned this when we were grooming the story. Changing the generator would not be backwards compatible in terms of reproducibility, so it would be appropriate to increment to version 2 when that change happens. RandomState is now legacy and not getting updated except when necessary because of an overall change to numpy, so to leverage new features and improvements we do eventually need to move to the new generator.

I see that you did update the example custom bandit. Thanks for that! My question is that since we designed this to be extensible, and BaseMAB is public for that purpose, should the create_rng function also be public since anyone writing a custom bandit will have to use it. I'll leave that up to you to decide.

skadio:
> @skadio I mentioned this when we were grooming the story. Changing the generator would not be backwards compatible in terms of reproducibility, so it would be appropriate to increment to version 2 when that change happens. RandomState is now legacy and not getting updated except when necessary to incorporate an overall change to numpy, so to leverage new features and improvements we do eventually need to move to the new generator.

Exactly! Thanks for raising all of this in the first place. We need to do the jump at some point, so let's see when there will be an opportunity/bandwidth. 

As for making create_rng() I can be persuaded each way. The first you mentioned it, I was yes.. later last night I was a no .. Today, I am more of a yes... :D I guess what's bothering about Customized Bandits is the split between somethings being public (like BaseMab to derive from)  and somethings private (like _Greedy to override). So which camp create_rng() go.. 

One way I can cast some logic into this is... BaseMAB is public and RNG is a parameter of BaseMAB... so rng creator to be public makes sense. That way, overall, BaseMAB+RNG remains public and the internal bandits remain private. 

Not the strongest design.. but I can live with that. So let me follow your suggestion and make it public.

PR #26, opened 2020-08-11
PyPi Release

erstrong:
Updates for release on PyPi.org

Signed-off-by: Emily Strong <emily.strong@fmr.com>


dorukkilitcioglu:
Regarding the changes - I feel like we can remove the wheel in the `dist` folder since it gets used automatically when you do `pip install`. That's what I'm aiming to do in my PR. Any thoughts?

erstrong:
@dorukkilitcioglu 
> Regarding the changes - I feel like we can remove the wheel in the `dist` folder since it gets used automatically when you do `pip install`. That's what I'm aiming to do for textwiser in my PR. Any thoughts?

It's something we can explore in the future, but for now we need to keep version 1.9 available and PiPi only has 1.10

dorukkilitcioglu:
As far as I can see the `dist` folder also only has version 1.10 - was 1.9 accidentally deleted? Right now you would have to go back & check out 1.9 from commit history to be able to install 1.9, which you can still do if we delete the `dist` folder in this PR.

PS: It might also be worth to upload 1.9 to PyPI if that's the case.  I'm fairly certain that PyPI would correctly realize that it's an outdated wheel and still serve 1.10 by default.

erstrong:
@dorukkilitcioglu Now that you mention it, it's in the version1.9 branch so it wouldn't be impacted.

setup.py
```diff
@@ -3,14 +3,19 @@
...
 setuptools.setup(
     name="mabwiser",
     description="MABWiser: Parallelizable Contextual Multi-Armed Bandits Library",
     long_description=long_description,
+    long_description_content_type="text/markdown",
     version="1.10.0",
     author="FMR LLC",
     url="https://github.com/fidelity/mabwiser",
     packages=setuptools.find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
+    install_requires=required,
```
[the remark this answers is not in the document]
erstrong: I added this after the initial push to PyPi but forgot to push the version increment so what's on there would get the update. If you reinstall you will get version 10.1 which does have the requirements.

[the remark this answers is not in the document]
bkleyn: Confirmed that this works now.

erstrong:
> Missed version in line 36

Pushed the correction

docs/api.html
```diff
@@ -317,55 +301,51 @@
...
-</table>
+<dl class="field-list simple">
+<dt class="field-odd">Author</dt>
+<dd class="field-odd"><p>FMR LLC</p>
+</dd>
+<dt class="field-even">Email</dt>
+<dd class="field-even"><p><a class="reference external" href="mailto:mabwiser&#37;&#52;&#48;fmr&#46;com">mabwiser<span>&#64;</span>fmr<span>&#46;</span>com</a></p>
+</dd>
+<dt class="field-odd">Version</dt>
+<dd class="field-odd"><p>1.10.0 of June 22, 2020</p>
```
bkleyn, line 312: Not a big deal, but for some reason this doc didn't update correctly?

docs/api.html
```diff
@@ -1381,22 +1268,21 @@
...
```
bkleyn, line 1279: Same here

docs/api.html
erstrong: Sphinx only recompiles the api page if something in the docstrings changes

bkleyn: Got it.

skadio: (still wanna keep the mabwiser email?)

erstrong: @skadio I forgot to make the update for this version and PyPi doesn't allow you to update existing version files so we would have to increment to 1.10.2. I've got a few cleanup items to do another patch for soon, so the email will get updated in that, and for now we can keep this pr focused on the PyPi release.

PR #27, opened 2020-09-03
LSH Approximate Nearest Neighbors

erstrong:
This PR adds a locality-sensitive hashing approximate nearest neighbors algorithm to the neighborhood policies. The design is intended to make it easy to add other ANNs in the future. It includes a secondary implementation for the simulator similar to the other nearest neighbors simulator classes. 

Benchmarking simulations comparing Radius to LSH:
- Train size: 700k
- Test size: 100k
- 8 jobs, threading backend

Radius: 
- Train: 1 second
- Test: 28 minutes
- Total runtime: 28 minutes
- 6.7% empty neighborhoods

LSH:
- Train: 3.5 minutes
- Test: 19 minutes
- Total runtime: 22.5 minutes
- 0 empty neighborhoods


CHANGELOG.txt
```diff
@@ -2,6 +2,18 @@
 MABWiser CHANGELOG
 =====================
 
+-------------------------------------------------------------------------------
+September, 3, 2020 1.11.0
+-------------------------------------------------------------------------------
+major:
+- New Algorithm: LSH Approximate Nearest Neighbors
```
skadio, line 9: may be spell out LSH

README.md
```diff
@@ -46,6 +46,7 @@ Available Learning Policies:
 * Upper Confidence Bound (UCB1)
 
 Available Neighborhood Policies: 
+* LSH Nearest
```
skadio, line 49: let's keep this lex ordered

README.md
```diff
@@ -61,8 +62,6 @@ our [documentation](https://fidelity.github.io/mabwiser/installation.html).
 
 Please submit bug reports and feature requests as [Issues](https://github.com/fidelity/mabwiser/issues).
 
-For additional questions and feedback, please contact us at [mabwiser@fmr.com](mailto:mabwiser@fmr.com?subject=[Github]%20MABWiser%20Feedback).
```
skadio, line 64: : )

docs/_sources/index.rst.txt
```diff
@@ -66,14 +66,15 @@ Available Learning Policies:
 
 Available Neighborhood Policies:
 
+* LSH Nearest
```
skadio, line 69: lex order

docs/_sources/new_bandit.rst.txt
```diff
@@ -262,9 +262,9 @@ Preparing a pull request typically involves the following steps:
 
 a. Add a note about your changes in the CHANGELOG.txt.
 b. Update the library version. You can use a keyword search for "version" to make sure you cover all version fields.
-c. Update the README.md, in necessary.
+c. Update the README.md, if necessary.
 d. Update the documentation rst files under the /docsrc folder , if necessary.
-e. If you update any documentation, make sure to recompile the docs by running ``make github`` under the /docsrc folder.
+e. If you update any documentation, make sure to recompile the docs by running ``make github`` under the /docsrc folder. If you have mabwiser installed, you will need to uninstall it, as Sphinx first looks for an installed version before using the local package.
```
skadio, line 267: quick q: if we make an update to doc files do we need to uninstall the library and then run Sphinx?

examples/contextual_mab.py
```diff
@@ -150,3 +150,31 @@
 
 # Updating of the model with new arm
 clusters.add_arm(6)
+
+###############################################################
+# LSH Approximate Neighborhood Policy with UCB1 Learning Policy
+###############################################################
+
+# Radius contextual policy with radius equals to 5 and ucb1 learning with alpha 1.25
```
skadio, line 158: is this comment correct?

examples/contextual_mab.py
```diff
@@ -150,3 +150,31 @@
...
+ann = MAB(arms=ads,
+          learning_policy=LearningPolicy.UCB1(alpha=1.25),
+          neighborhood_policy=NeighborhoodPolicy.LSHNearest(n_dimensions=5, n_tables=5))
+
+# Learn from previous ads shown and revenues generated
+ann.fit(decisions=train_df['ad'], rewards=train_df['revenues'], contexts=train)
```
skadio, line 164: consistency question: in other examples what do we call these objects, "mab"? or say "greedy" or "radius" etc.

mabwiser/approximate.py
```diff
@@ -0,0 +1,170 @@
...
+            # If neighbors exist
+            if len(indices) > 0:
+                predictions[index] = self._get_nhood_predictions(lp, indices, row_2d, is_predict)
+            else:  # When there are no neighbors
+                predictions[index] = self._get_no_nhood_predictions(lp, is_predict)
+
+        return predictions
+
+
+class _LSHNearest(_ApproximateNeighbors):
```
skadio, line 99: I am kinda following the code, it looks very clean. In parallel, it would be great to have a quick "algorithm" session on how this works from a learning perspective. (could be a journal club, or one of the team meetings etc.)

mabwiser/mab.py
```diff
@@ -413,6 +411,68 @@ def _validate(self):
             check_true(self.n_clusters >= 2, ValueError("The number of clusters must be at least two."))
             check_true(isinstance(self.is_minibatch, bool), TypeError("The is_minibatch flag must be a boolean."))
 
+    class LSHNearest(NamedTuple):
+        """LSH Approximate Nearest Neighbors Policy.
```
skadio, line 415: Spell out the first "LSH"

mabwiser/mab.py
```diff
@@ -413,6 +411,68 @@ def _validate(self):
...
+        For the simhash, contexts are projected onto an n-dimensional plane and each dimension is evaluated for its
+        sign. This is converted to a base 2 integer used as the hash value to assign the context to a hash table. This
+        process is repeated for a specified number of hash tables, where each has a unique, randomly-generated plane.
+        To select the neighbors for a context, the hash value is calculated for each hash table and any contexts with
+        the same hashes are selected as the neighbors.
+
+        As with the radius or k value for other nearest neighbors algorithms, selecting the best number of dimensions
+        and tables requires tuning. For the dimensions, a good starting point is to use the log of the square root of
+        the number of rows in the training data. This will give you sqrt(n_rows) number of hashes. The number of
+        dimensions and number of tables have inverse effects on the number of empty neighborhoods and average
```
skadio, line 429: can you confirm: based on n_dim, and n_tables, if you increase them, it is less approx and more exact? 
What would be the extreme case look like to get "exact" nhoods (and sure if that's possible though)

mabwiser/mab.py
```diff
@@ -697,7 +764,11 @@ def __init__(self,
             # Do not use parallel fit or predict for Learning Policy when contextual
             lp.n_jobs = 1
 
-            if isinstance(neighborhood_policy, NeighborhoodPolicy.Clusters):
+            if isinstance(neighborhood_policy, NeighborhoodPolicy.LSHNearest):
```
skadio, line 767: keep the lex order

mabwiser/neighbors.py
```diff
@@ -81,6 +79,13 @@ def _predict_contexts(self, contexts: np.ndarray, is_predict: bool,
         """Abstract method to be implemented by child classes."""
         pass
 
+    def _binarize_ts_rewards(self, decisions, rewards):
```
skadio, line 82: nice!

mabwiser/neighbors.py
```diff
@@ -101,10 +118,9 @@ class _Radius(_Neighbors):
     def __init__(self, rng: _BaseRNG, arms: List[Arm], n_jobs: int, backend: Optional[str],
                  lp: Union[_EpsilonGreedy, _Linear, _Popularity, _Random, _Softmax, _ThompsonSampling, _UCB1],
                  radius: Num, metric: str, no_nhood_prob_of_arm=Optional[List]):
-        super().__init__(rng, arms, n_jobs, backend, lp, metric)
+        super().__init__(rng, arms, n_jobs, backend, lp, metric, no_nhood_prob_of_arm)
```
skadio, line 121: can you comment on this update? I think I saw it's mentioned somewhere but couldn't track

mabwiser/simulator.py
```diff
@@ -242,15 +247,25 @@ def _get_nhood_predictions(self, lp, row_2d, indices, is_predict):
 
             return prediction, {}, arm_to_stat
 
+    def _get_no_nhood_predictions(self, lp, is_predict):
```
skadio, line 250: could this be imported instead of duplicate?

tests/test_approximate.py
```diff
@@ -0,0 +1,489 @@
...
+
+        self.assertListEqual(arms, [3, 1])
+
+    def test_greedy0_d2_single_test(self):
+
+        arms, mab = self.predict(arms=[1, 2, 3, 4],
+                                 decisions=[1, 1, 1, 2, 2, 3, 3, 3, 3, 3],
+                                 rewards=[0, 1, 1, 0, 0, 0, 0, 1, 1, 1],
+                                 learning_policy=LearningPolicy.EpsilonGreedy(epsilon=0),
+                                 neighborhood_policy=NeighborhoodPolicy.Radius(2),
```
skadio, line 57: why radious tested in test_approx?.

tests/test_approximate.py
```diff
@@ -0,0 +1,489 @@
...
+                                 context_history=[[1, 1, 2, 3, 5], [1, 2, 1, 1, 1], [0, 0, 1, 0, 0],
+                                                  [0, 2, 2, 3, 5], [1, 3, 1, 1, 1], [0, 0, 0, 0, 0]],
+                                 contexts=contexts,
+                                 seed=12,
+                                 num_run=1,
+                                 is_predict=True)
+
+        hashes = mab._imp.get_context_hash(np.asarray(contexts), mab._imp.table_to_plane[0])
+        for h in hashes:
+            self.assertEqual(len(mab._imp.table_to_hash_to_index[0][h]), 0)
```
skadio, line 229: I like the idea of checking the internals of hashes for behavior. 

This seems to be the case for the case when there are no nhodds. 

Is there a similar internal checking test for the case where there are neighbors where they should be assigned together etc?

tests/test_approximate.py
```diff
@@ -0,0 +1,489 @@
+# -*- coding: utf-8 -*-
```
skadio, line 1: is this testing approx abstract class OR testing lsh? 
- Have you considered having test_approx and test_lsh separately? Would that make sense? 
- if new approx methods arrive, do we want them in this file?

tests/test_approximate.py
```diff
@@ -0,0 +1,489 @@
...
+                                 context_history=[[0, 1, 2, 3, 5], [1, 1, 1, 1, 1], [0, 0, 1, 0, 0],
+                                                  [0, 2, 2, 3, 5], [1, 3, 1, 1, 1], [0, 0, 0, 0, 0],
+                                                  [0, 1, 4, 3, 5], [0, 1, 2, 4, 5], [1, 2, 1, 1, 3],
+                                                  [0, 2, 1, 0, 0]],
+                                 contexts=[[0, 1, 2, 3, 5]],
+                                 seed=123456,
+                                 num_run=1,
+                                 is_predict=True)
+
+        self.assertEqual(arms, 3)
```
skadio, line 67: behavior wise, good the confirm the result of "predict" from the outside. 

Anything you could check internally that things are working with data structures these hashes are tracking etc?

skadio: (I don't mean to do that for "every" test, but may be one dedicated test for internals if we don't have it covered already)

tests/test_simulator.py
```diff
@@ -20,13 +20,18 @@
...
-           LearningPolicy.Popularity(), LearningPolicy.Random(),
-           LearningPolicy.UCB1(), LearningPolicy.ThompsonSampling(),
+           LearningPolicy.Popularity(),
+           LearningPolicy.Random(),
+           LearningPolicy.UCB1(),
+           LearningPolicy.ThompsonSampling(),
            LearningPolicy.Softmax()]
 
-    parametric = [LearningPolicy.LinUCB()]
+    parametric = [LearningPolicy.LinUCB(), LearningPolicy.LinTS()]
```
skadio, line 29: nice!

skadio (approved):
This is a very nice addition! Thanks for turning it around quickly. 

Look forward to learning more on the method itself.

docs/_sources/new_bandit.rst.txt
dorukkilitcioglu: It might be easier to make sure that mabwiser is installed in editable/dev mode (`pip install - e. `) so that the local package and the installed version are the same. That's what I do for my TextWiser dev environment.

mabwiser/approximate.py
```diff
@@ -0,0 +1,170 @@
...
+        # Binarize the rewards if using Thompson Sampling
+        if isinstance(self.lp, _ThompsonSampling) and self.lp.binarizer:
+            self.rewards = self._binarize_ts_rewards(decisions, rewards)
+        else:
+            self.rewards = rewards
+
+        # Fit hashes for each training context
+        self._fit_operation(contexts)
+
+    def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray,
```
dorukkilitcioglu, line 42: This is a little confusing to me. In this default partial_fit implementation, are we effectively retraining the whole model?

mabwiser/approximate.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Add more historical data for prediction
+        self.decisions = np.concatenate((self.decisions, decisions))
+        self.rewards = np.concatenate((self.rewards, rewards))
+        self.contexts = np.concatenate((self.contexts, contexts))
+
+        # Fit hashes for each training context
+        self._fit_operation(contexts)
+
+    def _get_neighbors(self, row_2d):
```
dorukkilitcioglu, line 57: Why not make these abstract methods using the ABC class? Would be more explicit.

mabwiser/mab.py
```diff
@@ -413,6 +411,68 @@ def _validate(self):
...
+        ----------
+        n_dimensions: int
+            The number of dimensions to use for each plane.
+            Integer value. Must be greater than zero.
+            Default value is 5.
+        n_tables: int
+            The number of hash tables.
+            Integer value. Must be greater than zero.
+            Default value is 3.
+        no_nhood_prob_of_arm: None or List
```
dorukkilitcioglu, line 442: Might be good to also explicitly mention that the probabilities are used to select an arm randomly (that's not very explicit from this explanation).

mabwiser/mab.py
dorukkilitcioglu: Agreed with Serdar, making the effect of these params explicit would be very helpful.

mabwiser/neighbors.py
```diff
@@ -21,10 +21,12 @@
 class _Neighbors(BaseMAB):
 
     def __init__(self, rng: _BaseRNG, arms: List[Arm], n_jobs: int, backend: Optional[str],
-                 lp: Union[_EpsilonGreedy, _Linear, _Popularity, _Random, _Softmax, _ThompsonSampling, _UCB1], metric: str):
+                 lp: Union[_EpsilonGreedy, _Linear, _Popularity, _Random, _Softmax, _ThompsonSampling, _UCB1],
+                 metric: str, no_nhood_prob_of_arm=None):
```
dorukkilitcioglu, line 25: For consistency, can you add type hints to this? `Optional[List[float]]` would probably work.

setup.py
```diff
@@ -1,4 +1,5 @@
 import setuptools
+from mabwiser._version import __author__, __version__
```
dorukkilitcioglu, line 2: IIRC setting a version like this isn't encouraged. As you mentioned before, depending on how your PYTHONPATH is configured this might import the already installed mabwiser, which will have an old version.

I have a [workaround](https://github.com/fidelity/textwiser/blob/master/setup.py#L10-L11) in TextWiser, but you might be able to find something cleaner than that.

dorukkilitcioglu (commented):
I like the changes in general. The separation between the abstract approximate implementation and LSH is a little confusing - maybe separating some of the tests and adding ABCs would help.

Another general comment is that we might want to start thinking about separating our documentation PRs and code PRs when the changes to documentation are too much. Having 95 individual files updated in a single PR makes it harder to review.

Thanks for the hard work!

mabwiser/approximate.py
```diff
@@ -0,0 +1,170 @@
...
+        for k in self.table_to_plane.keys():
+            hash_value = self.get_context_hash(row_2d, self.table_to_plane[k])
+            indices += self.table_to_hash_to_index[k][hash_value[0]]
+
+        return indices
+
+    @staticmethod
+    def get_context_hash(contexts, plane):
+        # Project rows onto plane and get signs
+        projection_signs = 1 * (np.dot(contexts, plane) > 0)
```
bkleyn, line 162: Does the method make specific assumptions about the mean/variance of the contexts?

mabwiser/neighbors.py
bkleyn, line 25: Did we only have the "no neighborhood" functionality in simulator?

bkleyn (approved):
Great work Emily. Minor comments added. Confirmed tests pass.

docs/_sources/new_bandit.rst.txt
erstrong: @dorukkilitcioglu good to know! I spent an hour last week trying to figure out why the docs weren't updating until I remembered this issue.

examples/contextual_mab.py
erstrong: Copy paste error :)

examples/contextual_mab.py
erstrong: For the examples we name the object greedy, radius, etc. I'll change it to lshnearest

mabwiser/mab.py
erstrong: Simhash uses an approximation of cosine similarity, so I don't think there is a way to guarantee that you can get the same results as with Radius/KNearest with a cosine similarity metric. Increasing n_dimensions decreases the number of collisions for a given hyperplane, so the neighbors are more "exact", while increasing n_tables increases the diversity of the planes used and so finds neighbors the others miss. One thing to note is that this is O(2^n_dimensions) so some imprecision in the neighbors needs to be acceptable, otherwise you would be better off using Radius or KNearest.

mabwiser/neighbors.py
erstrong: The LSH algorithm has the possibility of zero neighbors if the dimensions are too high and there aren't enough tables, so I moved the empty neighborhood functionality to the parent Neighbors class.

mabwiser/simulator.py
erstrong: Good catch!

tests/test_approximate.py
erstrong: Whoops! Apparently PyCharm's search functionality is case sensitive

tests/test_approximate.py
erstrong: Added

tests/test_approximate.py
erstrong: The _ApproximateNearest parent class code cannot run independently, similar to _Nearest (The test_nearest.py tests are for KNearest)

tests/test_approximate.py
erstrong: Added

docs/_sources/index.rst.txt
```diff
@@ -1,4 +1,4 @@
-MABWiser Contextual Multi-Armed Bandits
+`MABWiser Contextual Multi-Armed Bandits
```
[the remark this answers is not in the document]
erstrong: Stray key stroke I'm guessing

mabwiser/approximate.py
erstrong: For LSH Nearest it just adds the new contexts to the existing hash tables. But now that it's been pointed out, I realize I forgot to update the indexes for those contexts. So thanks for noticing this!

If an algorithm needs to be retrained from scratch for each partial fit (e.g. necessary with the annoy library) it can access the updated self.contexts.

setup.py
erstrong: Thanks for the heads up! It worked correctly when I tested it, but I'll try the workaround you have

docs/_sources/new_bandit.rst.txt
```diff
@@ -262,9 +262,9 @@ Preparing a pull request typically involves the following steps:
 
 a. Add a note about your changes in the CHANGELOG.txt.
 b. Update the library version. You can use a keyword search for "version" to make sure you cover all version fields.
```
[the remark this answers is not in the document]
erstrong: There is still a version value that needs to be updated in the docs config.py, I couldn't get it to import. Though I may try the workaround Doruk mentioned above to see if it also works for the docs config.

mabwiser/approximate.py
```diff
@@ -0,0 +1,170 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
erstrong: I have it separate so that if we add other ANNs the neighbors.py won't become unmanageably long. I'll see what can be refactored, the trouble is a lot of the algorithm-specific logic is nested inside the duplicate code.

mabwiser/mab.py
skadio: cool, thanks for the comments. You can incorporate some of that to the docs in the code, so that reader has a direction on how to turn the knobs  on these params between approx -- exact spectrum,.

tests/test_approximate.py
skadio: test_nearest testing _Nearest makes perfect sense to me. Shall this be test_lsh then,?

mabwiser/approximate.py
erstrong: No, it's just meant to approximate the cosine similarity.

mabwiser/neighbors.py
erstrong: It previously belonged to the Radius class. I moved it to Neighbors so that ApproximateNeighbors could inherit it.

CHANGELOG.txt
```diff
@@ -2,6 +2,18 @@
...
+- This neighborhood policy uses locality-sensitive hashing with random projection via the SimHash algorithm to create approximate neighborhoods.
+- More information: https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Random_projection
+
+minor:
+- Refactor of _Neighbors and _Radius classes to support reusability of empty neighborhood functionality
+- Switch to centralized version incrementing and use of the opensource@fmr.com email
```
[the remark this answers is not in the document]
erstrong: The copyright hasn't changed, just how it's set.

setup.py
skadio: (If it turns out too complicated/involved, might wanna consider fixing version update a separate PR than LSH) 

Could you also comment once this version changes are in place in some way, like this or Doruk's workaround, what's the suggested procedure for the next person bumping version? Business, as usual, is to browse a few files and update the hardcoded values manually (and forget some and to pray the reviewers catch it). What's the process now?

docs/_sources/new_bandit.rst.txt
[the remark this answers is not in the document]
skadio: Seems I asked the same question above : )

mabwiser/approximate.py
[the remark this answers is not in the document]
skadio: OR, from a devil's advocate perspective, one might suggest don't worry about the abstraction until unless we are adding the "second" approximate neighbors. If we never need that (which is possible..), we skip the effort to code/maintain the abstract class in the first place.

mabwiser/neighbors.py
skadio: If my memory is correct, you can have this functionality for TS? Did n't we do this for S2S linTS/LinUCB

skadio: I completely forgot "why" we did for the Radius class only back then... Part of me says there was a reason... But no memory of it --and I am supposed to lose some memory too going forward .. so ¯\_(ツ)_/¯

erstrong: The linear classes don't use neighborhoods so they wouldn't have a use for this

setup.py
erstrong: We now only need to increment the version number in _version.py

docs/_sources/new_bandit.rst.txt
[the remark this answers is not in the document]
erstrong: The workaround also works for the config file, so we now only need to increment in _version.py

erstrong:
Pushed updates:
1. bug fix for partial fitting
2. updated docstrings with more detail about hyperparameters
3. workaround for version imports in setup.py and conf.py
4. additional unit tests
5. refactoring of duplicate functionality

docsrc/conf.py
```diff
@@ -14,19 +14,22 @@
 #
 import os
 import sys
+
 sys.path.insert(0, os.path.abspath('.'))
 sys.path.append(os.path.join(os.path.dirname(__name__), '..'))
 
+with open(os.path.join('..', 'mabwiser', '_version.py')) as fp:
```
dorukkilitcioglu, line 21: Good idea to also use this logic in `conf.py`! I might do the same the next time I update TextWiser documentation.

mabwiser/mab.py
```diff
@@ -453,6 +451,73 @@ def _validate(self):
...
+            The probabilities associated with each arm. Used select random arm if a prediction context has no neighbors.
```
dorukkilitcioglu, line 488: Missing a "to" here :)

mabwiser/neighbors.py
```diff
@@ -92,6 +97,18 @@ def _get_nhood_predictions(self, lp, indices, row_2d, is_predict):
         else:
             return lp.predict_expectations(row_2d)
 
+    def _get_no_nhood_predictions(self, lp, is_predict):
+
+        if is_predict:
+            # if no_nhood_prob_of_arm is None, select a random int
+            # else, select a non-uniform random arm
+            # choice returns an array, hence get zero index
+            rand_int = lp.rng.choice(len(self.arms), size=1, p=self.no_nhood_prob_of_arm)[0]
```
dorukkilitcioglu, line 106: FYI, with single item arrays you can also use `array.item()`. No need to change anything.

dorukkilitcioglu (dismissed):
Nice improvements to the abstraction logic! Looks good to go to me.

mabwiser/neighbors.py
erstrong: Good to know!

PR #31, opened 2021-04-05
switch l2 and alpha

bkleyn:


erstrong (approved):
Looks good! Thanks for following up on this!

skadio (approved):

PR #32, opened 2021-04-14
Numpy scikit version fix

filip-michalsky:
Fixing warnings for np.int as this gets deprecated in numpy 1.20.0+ and making forward-looking dtype change in base_mab for numpy 1.20.0+

Specifying sklearn needs to be below 0.24.0 so that unit tests don't fail.

### DIFF
--- a/mabwiser/base_mab.py
+++ b/mabwiser/base_mab.py
@@ -191,7 +191,7 @@ class BaseMAB(metaclass=abc.ABCMeta):
         n_jobs = self._effective_jobs(n_contexts, self.n_jobs)
 
         # Partition contexts between jobs
-        n_contexts_per_job = np.full(n_jobs, n_contexts // n_jobs, dtype=np.int)
+        n_contexts_per_job = np.full(n_jobs, n_contexts // n_jobs, dtype=int)
         n_contexts_per_job[:n_contexts % n_jobs] += 1
         starts = np.cumsum(n_contexts_per_job)
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 joblib
-numpy
+numpy<1.20.0
 pandas
-scikit-learn>=0.22.0
+scikit-learn>=0.22.0,<0.24.0
 scipy
 seaborn>=0.9.0
--- a/README.md
+++ b/README.md
@@ -56,7 +56,6 @@ MABWiser is available to install as `pip install mabwiser`. It can also be insta
 
 Please submit bug reports and feature requests as [Issues](https://github.com/fidelity/mabwiser/issues).
 
-
 ## Citation
 
 If you use MABWiser in a publication, please cite it as:


requirements.txt
```diff
@@ -1,6 +1,6 @@
 joblib
-numpy
+numpy<1.20.0
 pandas
-scikit-learn>=0.22.0
+scikit-learn>=0.22.0,<0.24.0
```
skadio, line 4: remind me again what's wrong with v24? 

sklearn 24 has lots of goodies, so I wonder if it will be possible to make a complete jump to that newer version in mab, and across libs.

skadio (approved):
Discard my Approval. Leaving it to @erstrong  and @bkleyn

requirements.txt
```diff
@@ -1,6 +1,6 @@
 joblib
-numpy
+numpy<1.20.0
```
skadio, line 2: We can't use any new numpy with mabwiser?

skadio (approved):

requirements.txt
erstrong: @filip-michalsky why are you adding the <0.24 requirement? The change in StandardScaler is an issue for use cases, not for the library

requirements.txt
erstrong: @filip-michalsky is the functionality no longer available, or does it just raise a deprecation warning?

mabwiser/base_mab.py
```diff
@@ -191,7 +191,7 @@ def _partition_contexts(self, n_contexts: int):
         n_jobs = self._effective_jobs(n_contexts, self.n_jobs)
 
         # Partition contexts between jobs
-        n_contexts_per_job = np.full(n_jobs, n_contexts // n_jobs, dtype=np.int)
+        n_contexts_per_job = np.full(n_jobs, n_contexts // n_jobs, dtype=int)
```
erstrong, line 194: Isn't this in conflict with the added version requirement? And seems like it would cause backward compatibility issues (ie should be bundled into a version increment)

erstrong (changes requested):

requirements.txt
filip-michalsky: I was not able to successfully run unit tests until sklearn was bumped down below 0.24. Replicated issue on both on my machine and a C2C box

requirements.txt
filip-michalsky: it raises a deprecation warning.

mabwiser/base_mab.py
filip-michalsky: This is not in conflict, tests pass. numpy 1.20.0+ would raise a warning, which this change fixes. Additionally, until we are able to bump to sklearn 0.24.0+, numpy 1.20.0+ would raise this deprecation warning inside sklearn itself.

requirements.txt
filip-michalsky: namely, with sklearn 0.24.0+, 6 tests fail:
======================================================================
FAIL: test_fit_twice_thompson_thresholds (test_clusters.ClustersTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/shared/corpid/a649154/pr110447-ai_coe_pi_atlas/projects/sr/PRTCJP-756/libraries/mabwiser/tests/test_clusters.py", line 274, in test_fit_twice_thompson_thresholds
    self.assertListEqual(arms, [3, 4])
AssertionError: Lists differ: [3, 3] != [3, 4]

First differing element 1:
3
4

- [3, 3]
?     ^

+ [3, 4]
?     ^


======================================================================
FAIL: test_greedy0_n3 (test_clusters.ClustersTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/shared/corpid/a649154/pr110447-ai_coe_pi_atlas/projects/sr/PRTCJP-756/libraries/mabwiser/tests/test_clusters.py", line 105, in test_greedy0_n3
    self.assertListEqual(arms, [2, 2])
AssertionError: Lists differ: [2, 3] != [2, 2]

First differing element 1:
3
2

- [2, 3]
?     ^

+ [2, 2]
?     ^


======================================================================
FAIL: test_partial_fit_thompson_thresholds (test_clusters.ClustersTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/shared/corpid/a649154/pr110447-ai_coe_pi_atlas/projects/sr/PRTCJP-756/libraries/mabwiser/tests/test_clusters.py", line 234, in test_partial_fit_thompson_thresholds
    self.assertListEqual(arms, [3, 4])
AssertionError: Lists differ: [3, 3] != [3, 4]

First differing element 1:
3
4

- [3, 3]
?     ^

+ [3, 4]
?     ^


======================================================================
FAIL: test_UCB1_c2 (test_parallel.ParallelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/shared/corpid/a649154/pr110447-ai_coe_pi_atlas/projects/sr/PRTCJP-756/libraries/mabwiser/tests/test_parallel.py", line 239, in test_UCB1_c2
    self.assertEqual(arm,  [[3, 3, 3, 1, 1, 1, 3, 1, 3, 3] for _ in range(5)])
AssertionError: Lists differ: [[3, 3, 1, 1, 1, 1, 3, 1, 3, 3], [3, 3, 1, 1, 1, [106 chars], 3]] != [[3, 3, 3, 1, 1, 1, 3, 1, 3, 3], [3, 3, 3, 1, 1, [106 chars], 3]]

First differing element 0:
[3, 3, 1, 1, 1, 1, 3, 1, 3, 3]
[3, 3, 3, 1, 1, 1, 3, 1, 3, 3]

Diff is 645 characters long. Set self.maxDiff to None to see it.

======================================================================
FAIL: test_UCB1_c2_backend (test_parallel.ParallelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/shared/corpid/a649154/pr110447-ai_coe_pi_atlas/projects/sr/PRTCJP-756/libraries/mabwiser/tests/test_parallel.py", line 882, in test_UCB1_c2_backend
    self.assertEqual(arm, [[3, 3, 3, 1, 1, 1, 3, 1, 3, 3] for _ in range(5)])
AssertionError: Lists differ: [[3, 3, 1, 1, 1, 1, 3, 1, 3, 3], [3, 3, 1, 1, 1, [106 chars], 3]] != [[3, 3, 3, 1, 1, 1, 3, 1, 3, 3], [3, 3, 3, 1, 1, [106 chars], 3]]

First differing element 0:
[3, 3, 1, 1, 1, 1, 3, 1, 3, 3]
[3, 3, 3, 1, 1, 1, 3, 1, 3, 3]

Diff is 645 characters long. Set self.maxDiff to None to see it.

======================================================================
FAIL: test_thompson_n3 (test_parallel.ParallelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/shared/corpid/a649154/pr110447-ai_coe_pi_atlas/projects/sr/PRTCJP-756/libraries/mabwiser/tests/test_parallel.py", line 579, in test_thompson_n3
    self.assertListEqual(arms, [1, 1, 3, 1, 1, 1, 1, 2, 1, 2])
AssertionError: Lists differ: [2, 1, 3, 3, 3, 2, 2, 3, 2, 3] != [1, 1, 3, 1, 1, 1, 1, 2, 1, 2]

First differing element 0:
2
1

- [2, 1, 3, 3, 3, 2, 2, 3, 2, 3]
+ [1, 1, 3, 1, 1, 1, 1, 2, 1, 2]

----------------------------------------------------------------------
Ran 472 tests in 184.675s

FAILED (failures=6)

bkleyn: I also confirmed that the 6 tests fail when scikit-learn jumps to 0.24+. They all seem related to a change in behavior in KMeans. I could not find any hints in the scikit-learn [changelog](https://scikit-learn.org/stable/whats_new/v0.24.html#version-0-24-0) that would explain the change though. Since it does not break any code and is a reproducibility issue I don't think it's necessary to explicitly restrict the version, but will need to figure out what we do with the tests.

erstrong: Ok I was able to reproduce the issue, looks like something changed with KMeans clustering again

requirements.txt
erstrong: We don't need to have a version restriction for a deprecation warning, that just means with some later version of numpy it won't be compatible

requirements.txt
filip-michalsky: yeah, I thought restricting the version of `sklearn` is a quick fix for this and the only alternative is to figuring out what changed in sklearn under the hood ...

mabwiser/base_mab.py
erstrong: It looks like the np.type format is aliases (https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) which explains this. I didn't know we could directly use them this way!

requirements.txt
erstrong: Looks like they did a complete rewrite of KMeans: https://github.com/scikit-learn/scikit-learn/commit/769794249c2e361c15ae15f4de8adcc02976ae74

bkleyn: Well that would explain it.

erstrong:
Once we finalize what changes are necessary, please update the changelog and increment the version in mabwiser/_version.py to 1.11.2 since this does include a code change

requirements.txt
skadio: first of all, disclaimer: I am not close to this work so my comments might be off.. 

I am not comfortable with an extremely strong <1.2 requirement. That just forbids the entire library to benefit from any upcoming numpy version.. 

If possible, and if we really want to avoid depreciation warning, we shall fix the code instead and make the jump to new numpy versions. 

If they are deprecating something, I am sure there is a (better) alternative now. We shall uptake that, instead of restricting ourselves to the past, no?

requirements.txt
skadio: again, disclaimer: I am not close to this work so my comments might be off..

I am not comfortable with the < 2.4 requirements. If they rewrite the KMeans, well that breaks our code.. which is natural, and here the tests are failing.. I would rather we update the tests and make the jump to newer versions of the library, no? 

We don't have "corporate customers" to maintain behavior for. Let's use that luxury and uptake new stuff?

skadio (changes requested):

requirements.txt
filip-michalsky: Agreed @skadio ! The reason I proposed the version restrictions is to propose a quick solution - albeit updating things on our end with a version bump makes more sense! I will send an update to this PR.

filip-michalsky:
> Once we finalize what changes are necessary, please update the changelog and increment the version in mabwiser/_version.py to 1.11.2 since this does include a code change

Thank you, I will make the changes based on this discussion and push 1.11.2 version. Thanks!

requirements.txt
erstrong: Serdar, per the discussion below, the issue is that if we require sklearn < 0.24, and a user has numpy = 1.20, sklearn has the np.dytpe code that is raising the deprecation warnings

PR #33, opened 2021-04-15
version 1.1.2 update

filip-michalsky:
version 1.1.2: 
- Unit tests fixed for the library to be compatible with sklearn 0.24.0+ (which changed its KMeans implementation)
- Fixed np.int raising deprecation warning with numpy versions of 1.20+

Signed-off-by: Filip Michalsky <Filip.Michalsky@fmr.com>


CHANGELOG.txt
```diff
@@ -2,6 +2,13 @@
 MABWiser CHANGELOG
 =====================
 
+-------------------------------------------------------------------------------
+April, 15, 2021 1.11.2
+-------------------------------------------------------------------------------
+minor:
+- Updated dtype from np.int to int in base_mab.py line 294 to resolve deprecation warning from numpy>=1.20.0
```
erstrong, line 9: I like the specificity 😆

CHANGELOG.txt
```diff
@@ -2,6 +2,13 @@
...
+- Updated tests to reflect changes in KMeans update in sklearn>=0.24.0
```
erstrong, line 10: The change is to the sklearn version dependency, the unit tests are incidental to that. Suggested rephrasing: "Updated sklearn version dependency to >=0.24.0 because of changes in the KMeans implementation"

erstrong (approved):
Overall looks good, just a suggestion for the changelog

CHANGELOG.txt
filip-michalsky: thank you, updated.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 joblib
 numpy
 pandas
-scikit-learn>=0.22.0
+scikit-learn>=0.24.0
```
skadio, line 4: Not about this PR, but I am very curious how >= 0.24 will play with other libraries. It would be amazing if we can make a jump to the newer version, across the board.

skadio (approved):

bkleyn (approved):

CHANGELOG.txt
filip-michalsky: @skadio I will check all other libraries as well

PR #35, opened 2021-07-28
TreeBandit Neighborhood Policy

irmakbky:
This PR is not immediately ready for merging, but I'm opening it for visibility. It introduces a new neighborhood policy, TreeBandit, that uses the [DecisionTreeClassifier() structure from sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html).


irmakbky:
Upon adding my TreeBandit example to examples/contextual_mab, I realized that the Kmeans policy gives an error saying it predicts [5, 2] instead of [1, 2]. @erstrong @bkleyn Can you check to see if that's also the case for you? We can change it if so.

filip-michalsky:
> Upon adding my TreeBandit example to examples/contextual_mab, I realized that the Kmeans policy gives an error saying it predicts [5, 2] instead of [1, 2]. @erstrong @bkleyn Can you check to see if that's also the case for you? We can change it if so.

>> Check your sklearn version. prior to 0.24, there was different KMeans implementation

irmakbky:
> Check your sklearn version. prior to 0.24, there was different KMeans implementation

It seems I'm using 0.24.2.

erstrong:
> > Check your sklearn version. prior to 0.24, there was different KMeans implementation
> 
> It seems I'm using 0.24.2.

That should be fine. We require >= 0.24.0. Which unit test is giving this error?

irmakbky:
> > > Check your sklearn version. prior to 0.24, there was different KMeans implementation
> > 
> > 
> > It seems I'm using 0.24.2.
> 
> That should be fine. We require >= 0.24.0. Which unit test is giving this error?

This is in examples/contextual_mab.py, [this line](https://github.com/fidelity/mabwiser/blob/c22b98e94da5728d7230600541c27ada967b8d90/examples/contextual_mab.py#L118) I think should assert [5, 2] instead.

erstrong:
> > > > Check your sklearn version. prior to 0.24, there was different KMeans implementation
> > > 
> > > 
> > > It seems I'm using 0.24.2.
> > 
> > 
> > That should be fine. We require >= 0.24.0. Which unit test is giving this error?
> 
> This is in examples/contextual_mab.py, [this line](https://github.com/fidelity/mabwiser/blob/c22b98e94da5728d7230600541c27ada967b8d90/examples/contextual_mab.py#L118) I think should assert [5, 2] instead.

Ah, looks like we missed that when we updated to require sklearn 0.24. Feel free to include the correction, or we can update it separately :)

filip-michalsky:
@irmakbky I can confirm that I also now get 3 unit test failures.

FAIL: test_partial_fit (test_treebandit.TreeBanditTest)
FAIL: test_treebandit (test_treebandit.TreeBanditTest)
FAIL: test_treebandit_expectations (test_treebandit.TreeBanditTest)

irmakbky:
> @irmakbky I can confirm that I also now get 3 unit test failures.
> 
> FAIL: test_partial_fit (test_treebandit.TreeBanditTest)
> FAIL: test_treebandit (test_treebandit.TreeBanditTest)
> FAIL: test_treebandit_expectations (test_treebandit.TreeBanditTest)

I think those tests are from the feature/TreeBandit branch which was the old one. Is it possible you ran those instead of the tests in tree-neighbor? There shouldn't be any unit tests with those names in this branch (tree-neighbor).

tests/test_treebandit.py
```diff
@@ -0,0 +1,348 @@
...
+                                                 [0, 2, 1, 0, 0]],
+                                contexts=[[0, 1, 2, 3, 5]],
+                                seed=123456,
+                                num_run=1,
+                                is_predict=True)
+
+        self.assertEqual(arm, 1)
+
+    def test_greedy1(self):
+        arms, mab = self.predict(arms=[1, 2, 3, 4],
```
filip-michalsky, line 82: @irmakbky curious how you came up with what values to assert. Pen and paper? These tests seem to be a bit more complicated from a dummy set to just "see " what it should equal to.

tests/test_treebandit.py
```diff
@@ -0,0 +1,348 @@
+# -*- coding: utf-8 -*-
+
+from mabwiser.mab import LearningPolicy, NeighborhoodPolicy
+from tests.test_base import BaseTest
+
+
+class TreeBanditTest(BaseTest):
```
filip-michalsky, line 7: a nice comprehensive set of tests!

filip-michalsky:
> > @irmakbky I can confirm that I also now get 3 unit test failures.
> > FAIL: test_partial_fit (test_treebandit.TreeBanditTest)
> > FAIL: test_treebandit (test_treebandit.TreeBanditTest)
> > FAIL: test_treebandit_expectations (test_treebandit.TreeBanditTest)
> 
> I think those tests are from the feature/TreeBandit branch which was the old one. Is it possible you ran those instead of the tests in tree-neighbor? There shouldn't be any unit tests with those names in this branch (tree-neighbor).

Ah you are right, I was on a wrong branch, have not noticed you changed it, my bad!

tests/test_treebandit.py
irmakbky: With learning policies like greedy with epsilon=0, since the context we're trying to predict for has been observed before, we know that in the trees, we'll end up reaching the same leaves. So, we expect it to predict the arm that gave the highest reward in the training. Although, now looking at it, maybe I should add more training data for the same context to test among the arm predictions better!

For other learning policies with exploration, I've had to observe the prediction first and then assert it. So, we're only checking if it predicts the same thing each time.

CHANGELOG.txt
```diff
@@ -2,6 +2,14 @@
 MABWiser CHANGELOG
 =====================
 
+-------------------------------------------------------------------------------
+August, 01, 2021 1.12.0
+-------------------------------------------------------------------------------
+major:
+- New Algorithm: TreeBandit as a decision-tree based neighborhood policy.
+- Many thanks to @irmakbky for her contribution!
```
skadio, line 10: Thank you heaps Irmak!! This is a great win for MABWiser!

examples/contextual_mab.py
```diff
@@ -115,7 +115,7 @@
 
 # Results
 print("KMeans: ", prediction, " ", expectations)
-assert(prediction == [1, 2])
+assert(prediction == [5, 2])
```
skadio, line 118: thanks for finding a bug in our examples ;)

mabwiser/mab.py
```diff
@@ -569,6 +571,48 @@ def _validate(self):
...
+        tree_parameters: Dict, **kwarg
+            Parameters of the decision tree.
+            The keys must match the parameters of sklearn.tree.DecisionTreeClassifier.
+            When a parameter is not given, the default of that parameter in sklearn.tree.DecisionTreeClassifier
+            will be chosen.
+            Default value is an empty dictionary.
+
+        Example
+        -------
+            >>> from mabwiser.mab import MAB, LearningPolicy, NeighborhoodPolicy
```
skadio, line 596: I assume you confirmed that this example works as-is and selects Arm2?

mabwiser/mab.py
```diff
@@ -569,6 +571,48 @@ def _validate(self):
...
+            >>> mab.predict([[3, 2, 0, 1]])
+            'Arm2'
+
+        """
+        tree_parameters: Dict = {}
+
+        def _validate(self):
+            check_true(isinstance(self.tree_parameters, dict), TypeError("tree_parameters must be a dictionary."))
+            tree = DecisionTreeClassifier()
+            for key in self.tree_parameters.keys():
```
skadio, line 612: ah that's a cool test!

mabwiser/treebandit.py
```diff
@@ -0,0 +1,163 @@
...
+        super().__init__(rng, arms, n_jobs, backend)
+        self.lp = lp
+        self.tree_parameters = tree_parameters
+        self.tree_parameters["random_state"] = rng.seed
+
+        self.arm_to_tree = None
+        self.arm_to_rewards = None
+
+        # Initialize the arm expectations to nan
+        # When there are neighbors, expectations of the underlying learning policy is used
```
skadio, line 34: not sure what does this comment mean?

mabwiser/treebandit.py
```diff
@@ -0,0 +1,163 @@
...
+        # When there are no neighbors, return nan expectations
+        reset(self.arm_to_expectation, np.nan)
+
+    def fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray = None) -> NoReturn:
+
+        # Reset the decision trees of each arm
+        self.arm_to_tree = {arm: DecisionTreeClassifier(**self.tree_parameters) for arm in self.arms}
+        self.arm_to_rewards = {arm: dict() for arm in self.arms}
+
+        if isinstance(self.lp, _ThompsonSampling) and self.lp.binarizer:
```
skadio, line 44: good call on binarizing these for TS!

mabwiser/treebandit.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Calculate fit
+        self._parallel_fit(decisions, rewards, contexts)
+
+    def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray = None) -> NoReturn:
+
+        # Update rewards list at leaves
+        for i, arm in enumerate(decisions):
+            leaf_index = self.arm_to_tree[arm].apply([contexts[i]])[0]
+            self.arm_to_rewards[arm][leaf_index] = np.append(self.arm_to_rewards[arm][leaf_index], rewards[i])
```
skadio, line 57: what is the data structure for arm_to_rewards[arm][index] .. is it a List or numpy Array? 

Cannot we do self.arm_to_rewards[arm][leaf_index].append(rewards[i])?

mabwiser/treebandit.py
```diff
@@ -0,0 +1,163 @@
...
+        arm_contexts = contexts[decisions == arm]
+        arm_rewards = rewards[decisions == arm]
+
+        # Check that the dataset for the given arm is not empty
+        if arm_contexts.size != 0:
+
+            # Fit decision tree of the arm with this dataset
+            self.arm_to_tree[arm].fit(arm_contexts, arm_rewards)
+
+            # For each leaf, keep a list of rewards
```
skadio, line 79: nice detailed comment!

mabwiser/treebandit.py
```diff
@@ -0,0 +1,163 @@
...
+
+            if is_predict:
+                predictions[index] = argmax(arm_to_expectation)
+            else:
+                predictions[index] = arm_to_expectation.copy()
+
+        # Return list of predictions
+        return predictions
+
+    def _uptake_new_arm(self, arm: Arm, binarizer: Callable = None, scaler: Callable = None):
```
skadio, line 139: do you have a corresponding unit test, where you start a TreeBandit and then add a new arm to it? 

Preferably, if you have a verified behavior in a multi-arm scenario, you can drop one arm, and build that scenario again starting with less arms and then adding that arm. The results should match?

mabwiser/treebandit.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.lp.add_arm(arm, binarizer)
+        self.arm_to_tree[arm] = DecisionTreeClassifier(**self.tree_parameters)
+        self.arm_to_rewards[arm] = dict()
+
+    def _get_leaf_lp(self, arm: Arm):
+
+        # TODO: maybe we don't need re-create but just change self.lp.arms to [arm, -1]?
```
skadio, line 147: did you get a chance to see, if simply revising the arms of the original MAB.lp object here produces identical results? 

If so, maybe we don't need to allocated a new LP for each region and instead, all leaf regions uses the "same" LP object that comes from MAB.LP.. which would be self.lp here. 

Of course, creating new objects each time seems "safer" but I wonder if the other is semantically equivalent..

tests/test_treebandit.py
skadio: excellent!

skadio (approved):
Done with my pass -- this looks in great shape! 

Please delete the other dangling branch, so we are not confused going forward.

CHANGELOG.txt
irmakbky: Thank you!

filip-michalsky (approved):
great work @irmakbky !

mabwiser/mab.py
irmakbky: I'll actually change its learning policy now to greedy epsilon=0 instead for a better test, but yes it works!

mabwiser/treebandit.py
irmakbky: I put that comment and initialization to be consistent with the other neighborhood policies. What it suggests is to initialize arm_to_expectation values to nan for the nhpolicy and we'll use the arm_to_expectation from the learning policy.

tests/test_treebandit.py
erstrong: In this test epsilon=1, ie there should be 100% randomization. I'm a little concerned that it's giving identical results to when epsilon=0. Could you add a second test with a different random seed to make sure that there isn't something else going on?

erstrong: (increasing the number of test contexts would would also work, eg predict 4 rather than 2)

mabwiser/treebandit.py
irmakbky: I meant for them to be np arrays because later when we're using indexing like arm_rewards[leaf_indices == index], I think it only works with np arrays. And I don't think just .append() to works with np arrays.

mabwiser/mab.py
```diff
@@ -569,6 +571,48 @@ def _validate(self):
...
                            ValueError("no_nhood_prob_of_arm should sum up to 1.0"))
 
+    class TreeBandit(NamedTuple):
+        """Tree Bandit Neighborhood Policy.
+
+        This policy fits a decision tree for each arm using context history.
+        It uses the leaves of these trees to partition the context space into regions
+        and keeps a list of rewards for each leaf.
+        To predict, it receives a context vector and goes to the corresponding
+        leaf at each arm's tree and applies the given MAB learning policy to select a reward
```
erstrong, line 581: This should specify that it uses context-free MAB lps

mabwiser/treebandit.py
irmakbky: test_add_arm tests adding the new arm, but I can add more test cases to check matching the results.

tests/test_invalid.py
```diff
@@ -293,6 +293,14 @@ def test_invalid_clusters_num(self):
         with self.assertRaises(ValueError):
             MAB([0, 1], LearningPolicy.EpsilonGreedy(epsilon=0), NeighborhoodPolicy.Clusters(n_clusters=1))
 
+    def test_invalid_treebandit_lp_linucb(self):
```
erstrong, line 296: Should there also be a test to validate that if unsupported arguments for DecisionTreeClassifier's parameters are given it will raise an error? Or are we leaving that up to sklearn?

irmakbky: _validate for TreeBandit does that I believe. Is that what you were referring to?

tests/test_treebandit.py
irmakbky: Makes sense! I've made it so that test_greedy0 and test_greedy1 have the same example but guess different arms for the first context. Is this ok?

mabwiser/treebandit.py
irmakbky: I checked and it unfortunately doesn't work. Changing the original lp creates problems, so it's probably better to keep _get_leaf_lp().

mabwiser/treebandit.py
irmakbky: I've realized a problem with this. If we want to add a new arm, we can add it, but we cannot partial_fit new context for that arm since TreeBandit doesn't do traditional partial fitting. So actually the issue here is that our partial_fit function only updates the leaves' list rewards, but it can't update for an arm that hasn't received any context/doesn't have a fitted tree yet.

skadio: Wow, that's subtle --great catch! 

In that case, can we plz raise a NotImplementedError in this _uptake_new_arm(). Add a test to verify that this operation is not supported, yet. 

Add your explanation/comment on why this is not implemented in this function.

skadio (approved):
@irmakbky  I can approve this PR so that you can merge it. Please proceed with a merge when you feel the reviewer's comments are addressed. It looks good from our end, and we can take care of the following edits if any are required. 

Great way to finish the cross-line with concrete impact! Congrats!! 🥳 🎉

mabwiser/treebandit.py
irmakbky: I added some code instead that tries to partial fit and if it realizes that the tree has not yet been fitted, it fits that specific tree with the given context. Hope that's fine!

mabwiser/treebandit.py
```diff
@@ -0,0 +1,171 @@
...
+        for i, arm in enumerate(decisions):
+            try:
+                if arm not in unfitted_arms:
+                    leaf_index = self.arm_to_tree[arm].apply([contexts[i]])[0]
+                    self.arm_to_rewards[arm][leaf_index] = np.append(self.arm_to_rewards[arm][leaf_index], rewards[i])
+            except AttributeError:
+                # If arm's tree has not been fitted
+                self.arm_to_tree[arm].fit(contexts[decisions == arm], rewards[decisions == arm])
+                unfitted_arms.append(arm)
+
```
skadio, line 66: minor: dangling extra blank line

tests/test_invalid.py
erstrong: Yes - checking that the keys are valid

mabwiser/treebandit.py
erstrong: I like that approach, it's what I would expect as an end-user to happen

skadio: Let me comment on this in the partial_fit() side

mabwiser/treebandit.py
```diff
@@ -0,0 +1,171 @@
...
+            self.lp.is_contextual_binarized = True
+
+        # Calculate fit
+        self._parallel_fit(decisions, rewards, contexts)
+
+    def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray = None) -> NoReturn:
+
+        # Update rewards list at leaves
+        # Keep track of arms whose trees have not been fitted, and fit them separately
+        unfitted_arms = []
```
skadio, line 57: You might not need to track of untrained arms. Technically, isn't it equivalent to rewards of that arm being empty. 

Alternatively, cannot we do this: 

- Loop thru arms, instead of decisions, as in the fit_arm
- Grab all the indices that matters for the current arm in one go
- Then, if (arm_to_reward[arm]) succeeds, as in the if-check in fit_arm(), we know that the arm is trained, we append the rewards as done here
- If the arm has no rewards recorded, (which can happen a) in the original fit() there was no data for this arm OR b) this is a new arm coming from _update_new_arm()), then first "fit()" the tree and then "save" its rewards.

erstrong:
A couple dangling required updates:
1. test_simulator.py has its own set of policy lists separate from the ones in test_base.py. This new policy needs to be added (and confirm the simulator unit tests pass with the new policy). I don't see any red flags for things that need to be customized in the simulator so you should be fine for that
2. Add the new policy to the list in docsrc/index.rst
3. Increment the version in _version.py

mabwiser/treebandit.py
irmakbky: Yes, exactly! I thought of this too, but then remembered that this was our initial approach and we changed it to its current version because it'd be more efficient. Do we want to change it back?

mabwiser/treebandit.py
```diff
@@ -0,0 +1,173 @@
...
+            # Each row needs a separately seeded rng for reproducibility in parallel
+            rng = np.random.RandomState(seed=seeds[index])
+
+            for arm in arms:
+
+                # Copy the row rng to the deep copied model in arm_to_tree
+                arm_to_tree[arm].rng = rng
+
+                # If there was prior data for this arm, get expectation for arm
+                if arm_to_rewards[arm]:
```
erstrong, line 124: The dictionary is created from self.arms, not from the set of arms that have training data. What is the behavior for an arm that doesn't have a fitted model?

mabwiser/treebandit.py
erstrong: What does the unfitted arms get used for? It doesn't seem to be available outside of this function

mabwiser/treebandit.py
```diff
@@ -0,0 +1,173 @@
...
+        # Initialize the arm expectations to nan
+        # When there are neighbors, in the leaf node of the tree, expectations of
+        # the underlying learning policy is used
+        # When there are no neighbors, return nan expectations
+        reset(self.arm_to_expectation, np.nan)
+
+    def fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray = None) -> NoReturn:
+
+        # Reset the decision trees of each arm
+        self.arm_to_tree = {arm: DecisionTreeClassifier(**self.tree_parameters) for arm in self.arms}
```
erstrong, line 42: So for every arm without training data there will be an unfitted DecisionTreeClassifier? Seems like it would be good to track those unfitted arms here as an attribute of the class (self.unfitted_arms) for later use

mabwiser/treebandit.py
skadio: Yeah, I wasn't sure how did we take care the untrained bit back then. 

One thing to consider; if sklearn.apply(X) allows finding the leaf index of multiple contexts at once, that would be quite nice actually, and the above alternative (as in the old way) would work?

Thoughts?

erstrong (changes requested):
Generally looks good but I'm concerned that I'm not seeing how the code handles arms that don't have training data. Specifically what happens to keep track of them, and how predictions can be made for them when using the context-free learning policies that support "cold" arms like epsilon greedy. 

Also missing a couple minor things noted in a previous comment.

irmakbky:
> A couple dangling required updates:
> 
> 1. test_simulator.py has its own set of policy lists separate from the ones in test_base.py. This new policy needs to be added (and confirm the simulator unit tests pass with the new policy). I don't see any red flags for things that need to be customized in the simulator so you should be fine for that
> 2. Add the new policy to the list in docsrc/index.rst
> 3. Increment the version in _version.py

Two questions:
1. test_simulator.py gives an error in [this line](https://github.com/fidelity/mabwiser/blob/ce6e7b37ae57b8398943a262d7060c7b9a193f3f/tests/test_simulator.py#L613) because it runs the parametric LinTS and LinUCB which are not to be run with TreeBandit since they are contextual. How should I avoid that error?
2. What exactly is the version I should increment it to?

skadio:
> > A couple dangling required updates:
> > 
> > 1. test_simulator.py has its own set of policy lists separate from the ones in test_base.py. This new policy needs to be added (and confirm the simulator unit tests pass with the new policy). I don't see any red flags for things that need to be customized in the simulator so you should be fine for that
> > 2. Add the new policy to the list in docsrc/index.rst
> > 3. Increment the version in _version.py
> 
> Two questions:
> 
> 1. test_simulator.py gives an error in [this line](https://github.com/fidelity/mabwiser/blob/ce6e7b37ae57b8398943a262d7060c7b9a193f3f/tests/test_simulator.py#L613) because it runs the parametric LinTS and LinUCB which are not to be run with TreeBandit since they are contextual. How should I avoid that error?
> 2. What exactly is the version I should increment it to?

1) I don't see good way other add a custom logic to skip LinTS/UCB when np is TreeBandit. Taht should be okay. 
2) Since this is a net-new algorithm you can bump it 1.12.0

mabwiser/treebandit.py
irmakbky: For the unfitted_arms, I can set it as a class attribute as @erstrong suggested and update it during _fit_arm and partial_fit and keep our current code. Or we can get rid of it and do it the way @skadio suggests. My only concern with the latter is that even though .apply() can work on multiple contexts, it'll end up giving us a list of leaves and we'll have to loop through them again. It might be cleaner to just keep an unfitted_arms list to keep track of when to update and when to fit from scratch.

mabwiser/treebandit.py
irmakbky: TreeBandit recognizes the arm and keeps an unfitted decision tree and a dictionary for leaf rewards for it. However, upon receiving a context to predict for, it likely wouldn't choose this arm since it's expected reward is 0.

erstrong:
> Two questions:
> 
> 1. test_simulator.py gives an error in [this line](https://github.com/fidelity/mabwiser/blob/ce6e7b37ae57b8398943a262d7060c7b9a193f3f/tests/test_simulator.py#L613) because it runs the parametric LinTS and LinUCB which are not to be run with TreeBandit since they are contextual. How should I avoid that error?
> 2. What exactly is the version I should increment it to?

1. you can add if logic to skip those combinations
2. 1.12.0 -- you have it in the changelog, but not in _version.py

mabwiser/treebandit.py
erstrong: In the fit function you call self.arm_to_rewards = {arm: dict() for arm in self.arms}. So 'arm' will always be available as a key in self.arm_to_rewards regardless of whether or not the arm has observed rewards and this line will always return true

mabwiser/treebandit.py
```diff
@@ -0,0 +1,175 @@
...
+                # If there was prior data for this arm, get expectation for arm
+                if arm_to_rewards[arm]:
+                    # Get leaf index for that context
+                    leaf_index = arm_to_tree[arm].apply([row])[0]
+
+                    # Get the rewards list of that leaf
+                    leaf_rewards = arm_to_rewards[arm][leaf_index]
+
+                    # Create leaf lp
+                    leaf_lp = self._get_leaf_lp(arm)
```
erstrong, line 134: So for every arm you are creating a new instance of the learning policy fit only to that arm?

mabwiser/treebandit.py
irmakbky: I just checked again, I believe when a numpy array is empty 'if nparray' evaluates to False.

mabwiser/treebandit.py
irmakbky: Yes. Specifically, we receive a context and follow it to a leaf in each arm's tree. Then, a new lp object is created for those leaves.

erstrong: So how does it create a softmax distribution or leverage epsilon for e-greedy?

irmakbky: Oh, I see your point! I guess all TreeBandit does is use the learning policy algorithms to choose an expected reward for each arm. But given that we're only using the lps on one arm at a time, some lps may not be used to their full potential.

mabwiser/treebandit.py
erstrong: I see

mabwiser/treebandit.py
```diff
@@ -0,0 +1,175 @@
...
+
+                    # Leaf LP: fit the same arm decision with the leaf rewards
+                    leaf_lp.fit(np.asarray([arm] * len(leaf_rewards)), leaf_rewards)
+
+                    # Leaf LP: predict expectation
+                    arm_to_expectation[arm] = leaf_lp.predict_expectations()[arm]
+
+            if is_predict:
+                predictions[index] = argmax(arm_to_expectation)
```
erstrong, line 143: This is purely greedy. The only policies where it maps 1:1 to how the policy predicts are Thompson Sampling and UCB1

erstrong: What I would recommend is have a single lp object for each context (initialized with the full set of arms) and partially fit on the leaf data for each arm. Then after looping through the arms, call predict or predict_expectations within the is_predict control logic

skadio:
@bkleyn @irmakbky There were a number of subtleties to take care of, but good to go from my end except for one thing:

We never talked about whether the underlying tree should be a Classification Tree or a Regression Tree. If I switch back and forth the unit tests are not impacted. Maybe that should be parameterized, maybe we can fix either one in the initial implementation.

bkleyn:
> We never talked about whether the underlying tree should be a Classification Tree or a Regression Tree...

Given that we do not constrain rewards to be binary, I think we have to use a Regression Tree. Probably fine to fix for now and can make configurable later if needed.

PR #39, opened 2022-02-14
NumPy RNG Update

dorukkilitcioglu:
This PR updates the NumPy RNG backend from the `RandomState` model to the new `Generator` model. This leads to a lot of test updates as well, since the random numbers also appear in tests.

At the same time, it updates the implementation of multivariate sampling, preferring to use NumPy's multivariate sampling with Cholesky method instead.

The PR is designated as WIP. Looking to get some reviews before updating the library version and the docs.

**Reviews:** Requesting review from @bkleyn to ensure the update to the backend and the update to the tests make sense.

Note that this does not remove `RandomState` from tests entirely, but from the library itself. That can be done in a separate PR.


mabwiser/utils.py
```diff
@@ -168,18 +168,46 @@ def standard_normal(self, size):
...
+
+        return mean + np.dot(sampled_norm, covar_decomposed)
+
 
 class _NumpyRNG(_BaseRNG):
 
     def __init__(self, seed):
         super().__init__(seed)
-        self.rng = np.random.RandomState(self.seed)
+        self.rng = np.random.default_rng(self.seed)
```
bkleyn, line 204: This will fail for Numy < 1.16?

bkleyn (approved):
Looks good to me.

Added a few minor questions. Most importantly it's not clear to me if change is functionally backwards compatible with older (<1.16) Numpy versions. 

Good to merge after final changes (CHANGELOG, version, etc.).

mabwiser/utils.py
```diff
@@ -168,18 +168,46 @@ def standard_normal(self, size):
         """
         pass
 
+    def multivariate_normal(self, mean, covariance, size=None):
```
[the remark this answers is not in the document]
dorukkilitcioglu: When I moved it here, I was thinking that this was a generic way of sampling from multivariate_normal, assuming we have a backend that implements `standard_normal`. As in, if we added another RNG backend, they don't _necessarily_ need to implement `multivariate_normal`, they only need to implement `standard_normal`.

I'm not sure if it's entirely necessary, the only usage I could see is if we implemented another RNG backend and that also had reproducibility issues with multivariate normal sampling. In which case, that backend should probably be responsible for implementing this code?

Let me know what you think, but I'm in favor of dropping it.

mabwiser/utils.py
dorukkilitcioglu: Correct, I'll update the reqs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
+name: ci
```
skadio, line 1: nice, nice, nice

mabwiser/linear.py
```diff
@@ -72,52 +72,19 @@ def _scale_predict_context(self, x):
 
 class _LinTS(_RidgeRegression):
 
-    def __init__(self, rng: _BaseRNG, l2_lambda: Num = 1.0, alpha: Num = 1.0,
```
skadio, line 75: ah nice! All this is gone, now? :)

mabwiser/treebandit.py
```diff
@@ -114,7 +114,7 @@ def _predict_contexts(self, contexts: np.ndarray, is_predict: bool,
         predictions = [None] * len(contexts)
         for index, row in enumerate(contexts):
             # Each row needs a separately seeded rng for reproducibility in parallel
-            rng = np.random.RandomState(seed=seeds[index])
```
skadio, line 117: @irmakbky thank you again for TreeBandits! :)

skadio (dismissed):
Looks good, just one comment about "cholesky" not being used as a parameter switch to multivariate_normal

mabwiser/utils.py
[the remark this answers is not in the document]
skadio: This threw me off a bit. 

I was expecting everything that touches RNG stuff to be in BaseRNG and/or in specific implementations  NumpyRNG. I see what you mean by this being a "utility" but still multivariate_normal() as a function related to RNG stuff

mabwiser/utils.py
```diff
@@ -168,18 +168,46 @@ def standard_normal(self, size):
...
+            -------
+            out : ndarray
+                A floating-point array of shape ``size`` of drawn samples
+        """
+        # Adapted from the implementation in numpy.random.generator.multivariate_normal version 1.18.0
+        # Uses the cholesky implementation from numpy.linalg instead of numpy.dual
+        sampled_norm = self.standard_normal(len(mean))
+
+        # Randomly sample coefficients from Normal Distribution N(mean=beta, std=covar_decomposed)
+        covar_decomposed = np.linalg.cholesky(covariance)
```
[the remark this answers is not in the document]
skadio: I see, thanks

bkleyn (approved):

README.md
```diff
@@ -1,3 +1,5 @@
+[![ci](https://github.com/fidelity/mabwiser/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/fidelity/mabwiser/actions/workflows/ci.yml)
```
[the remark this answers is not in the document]
dorukkilitcioglu: Done, customization can be done on readme sometime later.

dorukkilitcioglu:
@bkleyn I updated the library version, docs, and changelog. Also fixed some outdated example in the docs that was still reliant on `RandomState`. Let me know if there's anything missing or if we're good to go.

bkleyn (approved):

PR #50, opened 2022-03-03
Learning Policy Expectations

bkleyn:
Move policy-level logic to `predict_expectations` such that `predict` can directly use the predicted "expectations" to predict an arm. Specifically, each learning policy now predicts the arm with the highest value returned by `predict_expectations`.


mabwiser/popularity.py
```diff
@@ -33,9 +33,17 @@ def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.n
 
     def predict(self, contexts: np.ndarray = None) -> Arm:
 
-        # Select an arm randomized by expectation probability
+        # Return the first arm with maximum expectation
+        return argmax(self.predict_expectations())
+
+    def predict_expectations(self, contexts: np.ndarray = None) -> Dict[Arm, Num]:
+
+        # Return a random value between 0 and 1 for each arm that is "proportional" to the
```
dorukkilitcioglu, line 41: Huh, I think I understand why it isn't straightforward now. Does this create the same effect as random sampling based on the expectations? As in, if we continuously sampled from the previous implementation vs the current implementation, would we get a similar distribution of items sampled? As long as the _distribution_ of items as number of samples goes to infinity is the same as `self.arm_to_expectations`, I don't have a problem with the implementation.

dorukkilitcioglu: A bit of a sidenote, but I'm also wondering if we're cornering ourselves by calling this function `predict_expectations`. The "expectations" in this case _are_ static regardless of the context (as evidenced by `self.arm_to_expectation`). If the function was `predict_score`, then we could have a simpler implementation, or we wouldn't be bound by the expectations to match, just that the scores (and therefore arms) are ordered somehow.

To give an example, when we scored a context with the popularity policy before, it might've returned an item "a". The corresponding `predict_score` only needs to ensure that the score of "a" is highest. If you only had 2 items `a` and `b`, both `{a: 0.6, b:0.4}` and `{a:0.7, b:0.3}` would be valid results. In _expectation_, `b` still might get recommended more overall if `self.arm_to_expectations` had a higher value for `b`.

tests/test_popularity.py
```diff
@@ -27,9 +27,8 @@ def test_2arm_equal_prob(self):
                                 num_run=1,
                                 is_predict=False)
 
-        # Check that normalized probabilities are 1/2 each, and sum up to 1.0
```
dorukkilitcioglu, line 30: These aren't normalized now? Any reason why you wouldn't want to normalize the result of the dirichlet values?

dorukkilitcioglu (commented):
I think this is a huge step in reducing the discrepancy between learning policies. With the changes, `predict_expectations` is the defining part of the learning policies, while `predict` is just a utility function that calls argmax. Totally on board with this change.

I do have a couple of questions, mainly around the effect these changes have in terms of the distribution of recommendations for the affected learning policies. Have you done any experimentation on it? If so, can you tell a bit more?

dorukkilitcioglu (approved):

mabwiser/greedy.py
```diff
@@ -34,17 +34,17 @@ def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.n
 
     def predict(self, contexts: np.ndarray = None) -> Arm:
```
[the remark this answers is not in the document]
bkleyn: It's the same for all the non-contextual learning policies, but still differs for contextual and neighborhood policies. I can be convinced either way, but not a MUST in my opinion.

mabwiser/popularity.py
bkleyn: Re: *distribution* of items - Yes, I did a few experiments to confirm that the *distribution* of items are consistent. As expected, Random and Epsilon Greedy match very closely. Popularity and Softmax approach the distribution yielded by the previous implementation for very large number of samples.

bkleyn: Re: function name - I get what you are saying. To be fair this was already the case and I'm not sure if it warrants such a big change to the public API. Happy to chat more.

tests/test_popularity.py
bkleyn: They are already normalized (sum = 1).
https://numpy.org/doc/stable/reference/random/generated/numpy.random.dirichlet.html

mabwiser/popularity.py
```diff
@@ -33,9 +33,17 @@ def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.n
...
+        # expectation of the arm, by sampling from a Dirichlet distribution
```
filip-michalsky, line 42: Quick question: have we used Dirichlet for sampling before or is it part of the re-factor? I may not be as close to the mabwiser code base but this is the first time I am seeing it used.

bkleyn: It is the first time we're using in MABWiser.

mabwiser/greedy.py
```diff
@@ -34,17 +34,17 @@ def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.n
...
-        # Return the first arm with maximum expectation.
-        return argmax(self.arm_to_expectation)
+        # Return the first arm with maximum expectation
+        return argmax(self.predict_expectations())
 
     def predict_expectations(self, contexts: np.ndarray = None) -> Dict[Arm, Num]:
 
-        # Return a copy of expectations dictionary from arms (key) to expectations (values)
-        return self.arm_to_expectation.copy()
+        # Return a random expectation (between 0 and 1) for each arm with epsilon probability,
```
skadio, line 42: we thought about in the initial/past design but did not want to have a predict_exp() function that returns different outputs if you call it twice.. 

It seems we are OK with this behavior now?

mabwiser/popularity.py
skadio: Can you say a few words on why is popularity a Dirichlet distro and how is this diff than the multivariate dist we have already? Dir

mabwiser/rand.py
```diff
@@ -22,13 +22,13 @@ def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.n
 
     def predict(self, contexts: np.ndarray = None) -> Arm:
 
-        # Return a random arm
-        return self.arms[self.rng.randint(0, len(self.arms))]
+        # Return the first arm with maximum expectation
```
skadio, line 25: to be honest, I am not sold on this design. 

This might not be written somewhere as a contract, but one of the main/core ideas behind mabwiser was that mabwiser handles the arm_to_expectations for you.. 

Right now, in some policies, the calculation is done predict() and in some policies the calculation is done in predict_exp().. 

and now I am not even sure if I take a mab model and access to its arm_to_expectations anymore (is this even tracked .e.g, here) 

The ability to have access to arm_to_exp was nice, because then we could implement CustomLPs where on predict() we overwrite it and let mab handle arm_to_expectation and we could do problem-specific algorithms/decision on top of that. 

Is it still the case that I can trust "arm_to_expectations" of any mab object?

CHANGELOG.txt
```diff
@@ -2,6 +2,12 @@
 MABWiser CHANGELOG
 =====================
 
+-------------------------------------------------------------------------------
+March, 2, 2022 2.2.0
+-------------------------------------------------------------------------------
+major:
+- Modified `predict_expectations`, such that `predict` can use `predict_expectations` directly in all non-contextual learning policies
```
skadio, line 9: I am having a bit of a hard time understanding the contract. 

Is this the invariant now? That is "for any mab model, predict() calls predict_expectation() internally and then returns the decision? 

What happens to arm_to_expectations. Can I still access it, and have a guarantee that these values are correctly calculated/stored? 

Sepearate question, on which LP policies, predict OR predict_exp() is not deterministic? That is, I call it twice, I get a different result? Is this documented/shared with the user?

skadio (dismissed):
Wow, a lot of changes in Tests. Thank you for fixing! 

I have some comments, and I won't object to merging.

mabwiser/greedy.py
bkleyn: We already has this behavior. Note that `predict_expectations` for [ThompsonSampling](https://github.com/fidelity/mabwiser/blob/master/mabwiser/thompson.py#L57) would return different outputs if it is called twice.

mabwiser/rand.py
bkleyn: I don't agree with your summarization.

In this PR, for all learning policies the calculation is done in predict_expectations(), with predict() simply returning argmax(predict_expectations()).

In the current version there exists inconsistencies. For example, for Random the calculation happens in predict(), but for ThomsonSampling it happens in predict_expectations().

Re: ability to access `arm_to_expectations`. This has not been changed and can still be accessed.

What am I missing?

CHANGELOG.txt
bkleyn: All the non-contextual learning policies predict() calls predict_expectations() and then returns a decision. The implementation for [Linear](https://github.com/fidelity/mabwiser/blob/master/mabwiser/linear.py#L200) is a bit different, but it fundamentally does the same.

`arm_to_expectations` is still available and its definition has not changed.

Re: determinism - In the PR the only policies with deterministic predict() and predict_exp() output are LinUCB and UCB1. In the current version only LinUCB and UCB1 have deterministic predict(), whereas all the LP policies except TS and LinTS have deterministic predict_expectations().

mabwiser/popularity.py
bkleyn: For Popularity we calculate the average reward (assuming positive rewards) and then normalize it such that it sums to 1. We arm A has a value of 0.1 and arm B has a value of 0.2 (i..e 2x more popular), we want to predict B 2x more often than A. 

In the current version we achieve this using the [choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html) function, which randomly selects an arm according to the normalized values (weights). The Dirichlet distribution can be seen as a multivariate generalization of the Beta distribution and it can be used to sample the normalized values itself. I.e., for each arm it samples a random value between 0 and 1 in proportion to normalized values such that it sums to 1.

The multivariable normal distribution we have already assumes a normal distribution with a given mean and covariance. Values can be positive and negative and has no constraint wrt to the sum of the values.

skadio: Ah thanks! I was overlooking the "sum" detail

mabwiser/greedy.py
skadio: I see, makes sense

mabwiser/rand.py
skadio: GREAT to hear that arm_to_expectation remains intact. That was the main source of my confusion.

Can you help me understand this? Where does arm_to_exp is tracked? I don't see it under predict() or predict_exp()

CHANGELOG.txt
skadio: Gotcha, looks like determinism is off track either way.. that cannot be helped I guess. Doruk's paper should shed more light on this :D 

Re: arm_to_exp, good to hear it still is there.

mabwiser/rand.py
skadio: Ah forgot about my comment, looking into the code, I just realized that arm_to_exp is calculated in FIT and PARTIAL_FIT and has not much to do with PREDICT side of this. That's where I was confusing myself --that, we lost the arm_to_exp functionality. 

This resolves my comments.

mabwiser/popularity.py
skadio: Btw, it might help to incorporate some your comment here to the code comments about Dirichlet

filip-michalsky (dismissed):
Thanks for this addition, I like the streamline as well. I think some comments about using Dirichlet distribution for Popularity LP would be great for future reference.

Unrelated to this PR, I did look at the mabwiser public API - was the 'simulator' always there or is it a new addition?

mabwiser/greedy.py
dorukkilitcioglu: Re: determinism, I think there are ways to achieve it, such that consequent calls with the same `context` gives the same result (it would involve resetting the seed to _something_ each time `predict_expectations` is called). Not immediately necessary of course.

PR #52, opened 2022-03-08
Feature/lingreedy

wddcheng:


docs/_sources/new_bandit.rst.txt
```diff
@@ -274,6 +274,8 @@ Every test starts with the ``test_`` prefix followed by some descriptive name.
 To strengthen your test suite, consider other unittests with different number of arms,
 decisions and rewards to assert that your bandit behaves correctly.
 
+Add corresponding unittests in ``test_invalid.py`` to validate the parameters that are passed to your ``_MyCoolPolicy`` class.
```
bkleyn, line 277: Nice!

mabwiser/mab.py
```diff
@@ -190,6 +190,55 @@ def _validate(self):
...
+            >>> list_of_arms = ['Arm1', 'Arm2']
+            >>> decisions = ['Arm1', 'Arm1', 'Arm2', 'Arm1']
+            >>> rewards = [20, 17, 25, 9]
+            >>> contexts = [[0, 1, 2, 3], [1, 2, 3, 0], [2, 3, 1, 0], [3, 2, 1, 0]]
+            >>> mab = MAB(list_of_arms, LearningPolicy.LinGreedy(epsilon=0.5))
+            >>> mab.fit(decisions, rewards, contexts)
+            >>> mab.predict([[3, 2, 0, 1]])
+            'Arm2'
+        """
+        epsilon: Num = 0.05
```
bkleyn, line 230: You are using the same default epsilon as EpsilonGreedy? We've often used 10% as default, but this makes sense to me.

mabwiser/mab.py
```diff
@@ -812,12 +866,15 @@ def __init__(self,
             lp = _ThompsonSampling(self._rng, self.arms, self.n_jobs, self.backend, learning_policy.binarizer)
         elif isinstance(learning_policy, LearningPolicy.UCB1):
             lp = _UCB1(self._rng, self.arms, self.n_jobs, self.backend, learning_policy.alpha)
+        elif isinstance(learning_policy, LearningPolicy.LinGreedy):
+            lp = _Linear(self._rng, self.arms, self.n_jobs, self.backend, learning_policy.l2_lambda,
```
bkleyn, line 870: So `alpha` is explicitly set to zero for LinGreedy?

mabwiser/mab.py
```diff
@@ -812,12 +866,15 @@ def __init__(self,
...
+                         0, learning_policy.epsilon, "ridge", learning_policy.arm_to_scaler)
         elif isinstance(learning_policy, LearningPolicy.LinTS):
             lp = _Linear(self._rng, self.arms, self.n_jobs, self.backend, learning_policy.l2_lambda,
-                         learning_policy.alpha, "ts", learning_policy.arm_to_scaler)
+                         learning_policy.alpha, 0, "ts", learning_policy.arm_to_scaler)
```
bkleyn, line 874: Similarly, you are setting `epsilon` to zero for LinTS and LinUCB?

tests/test_lingreedy.py
```diff
@@ -0,0 +1,647 @@
+import datetime
+import math
+from copy import deepcopy
```
bkleyn, line 3: Pedantic comment. Add line between base and installed libraries.

tests/test_lingreedy.py
```diff
@@ -0,0 +1,647 @@
...
+import numpy as np
+import pandas as pd
+from sklearn.preprocessing import StandardScaler
+
+from mabwiser.mab import LearningPolicy, NeighborhoodPolicy
+from tests.test_base import BaseTest
+
+
+class LinGreedyTest(BaseTest):
```
bkleyn, line 12: Tests look good to me. Can you add one to confirm predictions for LinGreedy(epsilon=0, l2_lambda=X) == LinUCB(alpha=0, l2_lambda=X)?

bkleyn (changes requested):

mabwiser/mab.py
wddcheng: Yes, it's set to 0.05 to be consistent with EpsilonGreedy. I'm happy to hear what the team thinks about it.

mabwiser/mab.py
wddcheng: It is set to zero, although alpha is not used in LinGreedy anywhere.

mabwiser/mab.py
wddcheng: Yes, so that in predict_context() it will skip the random arm operations.

README.md
```diff
@@ -36,6 +36,7 @@ mab.predict()
 
 Available Learning Policies:
 * Epsilon Greedy [1, 2]
+* LinGreedy
```
skadio, line 39: you can refer to [1, 2] again

docs/about.html
```diff
@@ -4,7 +4,7 @@
   <meta charset="utf-8" /><meta name="generator" content="Docutils 0.17.1: http://docutils.sourceforge.net/" />
 
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-  <title>About Multi-Armed Bandits &mdash; MABWiser 2.2.0 documentation</title>
```
skadio, line 7: Unrelated to this PR --but I just noticed that in the Citations section the URL hyperrefs does not show up properly.

https://fidelity.github.io/mabwiser/#citation

Can we plz fix that as well, now that we are touching the docs?

bkleyn (approved):
Looks good to me, so I'm approving.

I'll let others chime in if there are remaining questions/comments.

bkleyn (approved):

skadio (dismissed):
Thank you @wddcheng !!

mabwiser/mab.py
```diff
@@ -881,10 +939,14 @@ def learning_policy(self):
             arm_to_scaler = dict()
             for arm in lp.arms:
                 arm_to_scaler[arm] = lp.arm_to_model[arm].scaler
-            if lp.regression == 'ts':
+            if lp.regression == 'ridge':
+                return LearningPolicy.LinGreedy(lp.alpha, lp.l2_lambda, arm_to_scaler)
```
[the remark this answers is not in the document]
bkleyn: No, you're right. `alpha` should be changed to `epsilon`. Thanks for catching!

mabwiser/linear.py
```diff
@@ -106,10 +106,12 @@ class _Linear(BaseMAB):
     factory = {"ts": _LinTS, "ucb": _LinUCB, "ridge": _RidgeRegression}
 
     def __init__(self, rng: _BaseRNG, arms: List[Arm], n_jobs: int, backend: Optional[str],
-                 l2_lambda: Num, alpha: Num, regression: str, arm_to_scaler: Optional[Dict[Arm, Callable]] = None):
+                 l2_lambda: Num, alpha: Num, epsilon: Num, regression: str,
```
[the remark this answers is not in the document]
bkleyn: Since all the linear policies require `l2_lambda` this ordering makes more sense to me. Is it worth changing API to LinUCB(l2_lambda, alpha, ...), LinTS(l2_lambda, alpha, ..) and LinGreedy(l2_lambda, epsilon)?

mabwiser/mab.py
bkleyn: Yes, I mean application-wise.

To be fair, I wouldn't say we have a true default for applications, with values typically ranging between 10% and 30%. So 10% would be closer to what we typically use, but I'm fine sticking with the 5%.

mabwiser/mab.py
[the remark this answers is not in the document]
wddcheng: Nice catch! @skadio 
now I'm confused why with this bug here the correct epsilon actually gets passed to the inner functions

mabwiser/linear.py
```diff
@@ -190,12 +192,19 @@ def _predict_contexts(self, contexts: np.ndarray, is_predict: bool,
...
+            # With epsilon probability set arm expectation to random value
+            if rng.rand() < self.epsilon:
+                for arm in arms:
+                    arm_to_expectation[arm] = rng.rand()
 
-                # Get the expectation of each arm from its trained model
-                arm_to_expectation[arm] = arm_to_model[arm].predict(row)
+            else:
+                # Create new seeded generator for model to ensure reproducibility
+                model_rng = create_rng(seed=seeds[index])
```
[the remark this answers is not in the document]
bkleyn: I suggested this so let me cover. 

No we don't need both. However, note that since we now use the generator (`rng`) in line 196, it will have a different state when the model generator is set in line 204. To preserve the same output we had before and avoid fixing many tests I suggested this.

Admittedly, it would be cleaner to use generator.

mabwiser/mab.py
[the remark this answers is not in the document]
wddcheng: that makes sense!

mabwiser/mab.py
```diff
@@ -190,6 +190,56 @@ def _validate(self):
...
+            >>> mab.fit(decisions, rewards, contexts)
+            >>> mab.predict([[3, 2, 0, 1]])
+            'Arm2'
+        """
+        epsilon: Num = 0.05
+        l2_lambda: Num = 1.0
+        arm_to_scaler: Dict[Arm, Callable] = None
+
+        def _validate(self):
+            print(self.epsilon)
```
bkleyn, line 235: Not sure when this snuck in. Please drop.

mabwiser/linear.py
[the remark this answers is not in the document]
bkleyn: Actually ignore what I said, I'm not sure if putting l2_lambda first makes more sense.

mabwiser/linear.py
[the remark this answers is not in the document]
bkleyn: Actually ignore what I said, I'm not sure if putting l2_lambda first makes more sense.

mabwiser/linear.py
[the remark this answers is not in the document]
skadio: @bkleyn  Just to confirm my understanding: "so we don't need both and we could have used the existing rng on L196 BUT due to state if we make this change, we have to fix soo many output in tests, hence we can afford sticking to this as-is" 

Am I right?

mabwiser/linear.py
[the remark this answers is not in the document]
wddcheng: Based on the discussion, so the sequence in the __init__ parameters would be alpha, epsilon, l2_lambda, regression, etc?

mabwiser/linear.py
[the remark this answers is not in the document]
bkleyn: Yes, that's correct.

mabwiser/mab.py
bkleyn: Actually I'm going to suggest changing it to 10%. As mentioned above, this would be closer to application-level defaults and 5% is almost pure exploiting.
@skadio - please chime in if you disagree.

mabwiser/linear.py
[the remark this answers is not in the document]
bkleyn: Yes, changing ordering to _Linear(... alpha, epsilon, l2_lambda, regression, ...) would be "more consistent" with public API I think.
@skadio - Do you have another suggestion?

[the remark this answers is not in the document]
skadio: What's the most common param for all? I would put that one first, and so-on. 

BUT if this leads to a change in the Public API (that is it would break others/users code in subtle ways in the next pip install (and they won't notice it since these params interoperate due to common 0-1 range --then I am not inclined to make any changes.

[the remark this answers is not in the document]
bkleyn: Makes sense. l2_lambda is the common param for all the linear policies, BUT putting that first will break public API.

@wddcheng - Let's just change internal _Linear to at least have partial ordering consistent with public API, as you suggested above.

mabwiser/mab.py
skadio: agreed!

wddcheng: and it's now changed!

bkleyn (approved):

mabwiser/mab.py
```diff
@@ -62,12 +62,61 @@ class EpsilonGreedy(NamedTuple):
...
+        epsilon: Num = 0.1
 
         def _validate(self):
             check_true(isinstance(self.epsilon, (int, float)), TypeError("Epsilon must be an integer or float."))
             check_true(0 <= self.epsilon <= 1, ValueError("The value of epsilon must be between 0 and 1."))
 
+    class LinGreedy(NamedTuple):
+        """LinGreedy Learning Policy.
+
+        This policy trains a ridge regression for each arm.
```
filip-michalsky, line 74: ridge regression 'model'?

mabwiser/mab.py
```diff
@@ -62,12 +62,61 @@ class EpsilonGreedy(NamedTuple):
...
+            Integer or float. Must be between 0 and 1.
+            Default value is 0.1.
+        l2_lambda: Num
+            The regularization strength.
+            Integer or float. Cannot be negative.
+            Default value is 1.0.
+        arm_to_scaler: Dict[Arm, Callable]
+            Standardize context features by arm.
+            Dictionary mapping each arm to a scaler object. It is assumed
+            that the scaler objects are already fit and will only be used
```
filip-michalsky, line 92: 'fitted'?

PR #53, opened 2022-03-17
Vectorized Predict for Non-contextual Policies

bkleyn:
* Vectorize the pred_exp() function for non-contextual policies
* To enable multi-decision predict, allow non-contextual policies to accept contexts input (e.g., empty list) to indicate that multiple prediction should be generated
* Type-hinting and PEP8 clean up


skadio (approved):
Blank approval, as I wont have time.

one quick question, how does specifying none contexts enables vectorization?

bkleyn:
> one quick question, how does specifying none contexts enables vectorization?

Predictions will only be vectorized when contexts is not none. In terms of usage it looks like this.

mab = MAB([1, 2, 3], LearningPolicy.Random())
mab.fit(decisions=[1, 1, 2, 2], rewards=[0, 1, 1, 1])
predictions = mab.predict([[], [], []])   # generate 3 predictions

mabwiser/mab.py
```diff
@@ -587,7 +587,7 @@ class Radius(NamedTuple):
             Accepts any of the metrics supported by scipy.spatial.distance.cdist.
             Default value is Euclidean distance.
         no_nhood_prob_of_arm: None or List
-            The probabilities associated with each arm. Used to select random arm if a prediction context has no neighbors.
+            The probabilities associated with each arm. Used to select random arm if context has no neighbors.
```
dorukkilitcioglu, line 590: `no_nhood_prob_of_arm` description is different between this line and line 540, `context` vs `prediction`.

mabwiser/greedy.py
```diff
@@ -32,19 +32,33 @@ def fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray =
...
+                                                                                   List[Dict[Arm, Num]]]:
 
         # Return a random expectation (between 0 and 1) for each arm with epsilon probability,
-        # and the actual arm expectations otherwise
-        if self.rng.rand() < self.epsilon:
-            return dict((arm, self.rng.rand()) for arm in self.arms).copy()
+        # and the actual arm expectations otherwise.
+        # If contexts is None or has length of 1 generate single arm to expectations,
+        # otherwise use vectorized functions to generate a list of arm to expectations with same length as contexts.
+        if contexts is None or len(contexts) == 1:
```
dorukkilitcioglu, line 51: We might need some input sanitization here.

If I gave a single user context as input (ex: [4, 3, -1]), how would greedy handle it vs a linear policy? Greedy would assume it is 3 different contexts and give 3 different results. I don't think that would be the intended result. What about linear?

If you agree that this isn't desired, perhaps we can validate whether `contexts` is a list or a list of lists (or a 2D array vs 1D array), and pick the behavior that way.

mabwiser/neighbors.py
```diff
@@ -64,12 +64,12 @@ def partial_fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.n
         self.contexts = np.concatenate((self.contexts, contexts))
         self.rewards = np.concatenate((self.rewards, rewards))
 
-    def predict(self, contexts: np.ndarray = None):
+    def predict(self, contexts: np.ndarray = None) -> Union[Arm, List[Arm]]:
```
dorukkilitcioglu, line 67: Thanks for all the type hints!

dorukkilitcioglu (changes requested):
Looks good overall, but I think the PR needs some clarification on how everything behaves with the rather large range of input formats. Specifically, I don't think giving a single context or a 1D numpy array to the non-contextual policies give the correct result (or let me know if it does and I'm just wrong). Once that's addressed I'm happy to approve.

dorukkilitcioglu (approved):
Learned that contexts have to be 2D arrays, so there's no reason for concern. G2G from my side.

mabwiser/greedy.py
bkleyn: Contexts have to be 2D so we're good.

PR #65, opened 2023-01-24
Add cold arm definition to MAB object

dorukkilitcioglu:
This PR adds cold arm set to the MAB object.


tests/test_mab.py
```diff
@@ -1767,3 +1767,25 @@ def test_pickle_fitted_reproducibile_expectations(self):
...
+                              seed=123456,
+                              num_run=4,
+                              is_predict=True)
+
+        # Before warm start
+        self.assertEqual(mab._imp.trained_arms, [1, 2])
+        self.assertSetEqual(mab.cold_arms, {3})
+
+        # Warm start
+        mab.warm_start(arm_to_features={1: [0, 1], 2: [0, 0], 3: [0.5, 0.5]}, distance_quantile=0.5)
```
skadio, line 1790: Am I reading this right: 

So we already have a "warm_start()" function to copy models based on features and quantile. (was that the other PR you had?) 

And now, here, you are adding the ability to for the system to store and track "cold_arms". 

Together, a user can "warm_start" things, and at any point, ask what is "cold". 

Design question; any pros/cons if this was not storing a set of cold_arms only but instead sth like "arm_to_status" where status can be "cold", "trained", "warm", "warm+tuned", etc.

mabwiser/_version.py
```diff
@@ -3,5 +3,5 @@
 
 __author__ = "FMR LLC"
 __email__ = "opensource@fmr.com"
-__version__ = "2.5.0"
+__version__ = "2.5.1"
```
skadio, line 6: since this is a new capability, looks like a 2.6 to me

skadio (dismissed):
LGTM, one learning question and one design question

mabwiser/_version.py
bkleyn: In my mind this is an extension of the previous PR and fundamentally covers the same warm_start() capability?

tests/test_mab.py
dorukkilitcioglu: As Bernard said, the previous PR was on updating warm start to account for multiple calls to `warm_start`.

I really like the `arm_to_status` idea! We could use that to consolidate `trained_arms` and `cold_arms`.

```python
arm_to_status = {"a": "cold", "b", "warm", "c": "trained"}
```

I'm a bit skeptical of also merging in `cold_arm_to_warm_arm`, because then we'd have to incorporate that info into the `arm_to_status` dict, which could make it a little more confusing. However, I really like the single source of truth with one object, so I'd prefer to implement that as well. Consider something like this:

```python
arm_to_status = {"a": "cold", "b": ("warm", "a"), "c": "trained"}

File: mabwiser/_version.py (hunk 2 above)
Comment (dorukkilitcioglu) [reply mabwiser/_version.py:6]: Exactly my reasoning

File: tests/test_mab.py (hunk 1 above)
Comment (skadio) [reply tests/test_mab.py:1790]: I am like the design considerations.. just wanted to mention cold, warm, trained, and warm+tuned (if possible). There will be two things critical 1) setting these correctly, and 2) (harder) to maintain them correctly going forward. For instance fit() operation now has to switch a warm arm to warm+tuned when it sees the first data point and so on. 

I am OK with "status" as the field name, as in, arm_to_status. I don't have any better suggestion. My only consideration is if/when that is used from outside, gotta make sure if it something that is hard to confuse with arm_to_is_active (like I did tripped last time)

File: mabwiser/_version.py (hunk 2 above)
Comment (skadio) [reply mabwiser/_version.py:6]: kinda disagree (and it could be because I am not close..) but if I am reading the PR right, it introduces a compeletely new field that was not there before; mabwiser.cold_arms OR as we brainstorm in the other thread mabwiser.arm_to_status.. 

My rationale is (IF) this is net new, then it's not an amendment/fix to something existing, hence not 2.5.1 and 2.6.0, but it does not break anything backward or change the paradigm going forward hence not 3.0.0 and 2.6.0

Comment (skadio) [reply mabwiser/_version.py:6]: your call, OK either way

File: mabwiser/base_mab.py [hunk 3]
```diff
@@ -83,8 +87,25 @@ def __init__(self, rng: _BaseRNG, arms: List[Arm], n_jobs: int, backend: str = N
...
+
+    @property
+    def trained_arms(self) -> List[Arm]:
+        """List of trained arms.
+
+        Arms for which at least one decision has been observed are deemed trained."""
+        return [arm for arm in self.arms if self.arm_to_status[arm][STATUS_TRAINED]]
+
+    @property
+    def cold_arms(self) -> List[Arm]:
```
dorukkilitcioglu, line 106: Quick caveat here: by this definition, all arms in the neighborhood policies are cold. At a LP level, this might make sense.. but I don't think it makes sense at MAB level, so the `cold_arms` property in the MAB object has additional logic to make all arms "warm" if it's using a neighborhood policy.

mabwiser/base_mab.py
```diff
@@ -394,3 +402,34 @@ def _get_cold_arm_to_warm_arm(self, cold_arm_to_warm_arm, arm_to_features, dista
...
+            # Get list of arms in decisions
+            arms = np.unique(decisions).tolist()
+
+        for arm in self.arms:
+            if arm in arms:
+                # All system arms are now trained
+                self.arm_to_status[arm][STATUS_TRAINED] = True
+
+                # If fitting from scratch, arm is no longer warm started
+                if not is_partial:
```
dorukkilitcioglu, line 433: Thru this logic, we can now tell when an arm has only been partial fit after a warm start.

mabwiser/base_mab.py
```diff
@@ -394,3 +403,34 @@ def _get_cold_arm_to_warm_arm(self, cold_arm_to_warm_arm, arm_to_features, dista
...
+    def _reset_arm_to_status(self):
+        self.arm_to_status: Dict[Arm, dict] = {arm: {STATUS_TRAINED: False, STATUS_WARM: False,
+                                                     STATUS_WARM_STARTED_BY: None}
+                                               for arm in self.arms}
+
+    def _set_arms_as_trained(self, arms: Optional[List[Arm]] = None, decisions: Optional[np.ndarray] = None,
+                             is_partial: bool = True):
+        """Sets the given arms as trained.
+
+        Uses the ``arms`` if provided, otherwise calculates the arms from the ``decisions``.
```
bkleyn, line 416: Do you ever use this with arms not None?

dorukkilitcioglu: No, since it's always calculated in LPs using the decisions. However, I could see a custom definition for a neighborhood policy or a new LP which isn't just unique arms in decisions.

mabwiser/_version.py
dorukkilitcioglu: With the arm to status changes, I've bumped to 2.6.0

mabwiser/base_mab.py
```diff
@@ -22,6 +22,10 @@
 __version__ = __version__
 __copyright__ = __copyright__
 
+STATUS_TRAINED = 'trained'
```
dorukkilitcioglu, line 25: Replace this with `is_trained`?

bkleyn: Yes, that's better. Same for `warm` to `is_warm`.

mabwiser/base_mab.py
bkleyn: Maybe remove that bit given that it's not used? Looking to remove some complexity and making it easier to maintain if/when we look at this again.

mabwiser/base_mab.py
dorukkilitcioglu: I'm not sure if it makes things any simpler for cold arms. We don't actually use the cold arms for neighborhood policies, and even if we replaced warm start with a simple `pass` in the neighborhood policy wouldn't we still need to have a definition of cold arms if we're making that available at MAB level?

bkleyn: No not for cold arms specifically. But it would be more consistent to say that warm start works for all LPs (in pretty much the same way), but that NPs cannot be warm started. If there are no major objections, I'm going to bundle this change here.

bkleyn:
@skadio I wrapped up the changes from first round review
- adding `arm_to_status` in BaseMAB that can be used to easily track if an arm is trained (arm was fit or partial fit), if an arm was warm started, and by what arm it was warm started
- restricted warm start to Learning Policies as it is not clear how to interpret cold arms for Neighborhood Policies

CHANGELOG.txt
```diff
@@ -2,6 +2,13 @@
 MABWiser CHANGELOG
 =====================
 
+January, 24, 2022 2.6.0
+-------------------------------------------------------------------------------
+minor:
```
skadio, line 7: As a complete new functionality (correct me if I am wrong) this seems major to me (as in, if I were  a user, I would wanna learn this)

mabwiser/base_mab.py
```diff
@@ -64,11 +68,11 @@ class BaseMAB(metaclass=abc.ABCMeta):
         Default value is None. In this case the default backend selected by joblib will be used.
     arm_to_expectation: Dict[Arm, float]
         The dictionary of arms (keys) to their expected rewards (values).
-    cold_arm_to_warm_arm: Dict[Arm, Arm]:
-        Mapping indicating what arm was used to warm-start cold arms.
-    trained_arms: List[Arm]
-        List of trained arms.
-        Arms for which at least one decision has been observed are deemed trained.
+    arm_to_status: Dict[Arm, dict]
+        The dictionary of arms (keys) to their status (values), where the status consists of
```
skadio, line 72: so what happens when an arm is warm_started and then trained on top?

mabwiser/base_mab.py
```diff
@@ -83,8 +87,26 @@ def __init__(self, rng: _BaseRNG, arms: List[Arm], n_jobs: int, backend: str = N
...
+    def trained_arms(self) -> List[Arm]:
+        """List of trained arms.
+
+        Arms for which at least one decision has been observed are deemed trained."""
+        return [arm for arm in self.arms if self.arm_to_status[arm][STATUS_TRAINED]]
+
+    @property
+    def cold_arms(self) -> List[Arm]:
+        """List of cold arms"""
+        return [arm for arm in self.arms if ((not self.arm_to_status[arm][STATUS_TRAINED]) and
```
skadio, line 108: (minor) can remove parantheses

mabwiser/base_mab.py
```diff
@@ -83,8 +87,26 @@ def __init__(self, rng: _BaseRNG, arms: List[Arm], n_jobs: int, backend: str = N
...
 
         self.arm_to_expectation: Dict[Arm, float] = dict.fromkeys(self.arms, 0)
-        self.cold_arm_to_warm_arm: Dict[Arm, Arm] = dict()
-        self.trained_arms: List[Arm] = list()
+        self._reset_arm_to_status()
+
+    @property
+    def cold_arm_to_warm_arm(self) -> Dict[Arm, Arm]:
+        """Mapping indicating what arm was used to warm-start cold arms."""
+        return {cold_arm: status[STATUS_WARM_STARTED_BY] for cold_arm, status in self.arm_to_status.items()
```
skadio, line 95: I cannot really parse the code a glance.. In the For loop, the if condition returns only warm arms, but then why is it called cold_arm?

mabwiser/base_mab.py
```diff
@@ -94,12 +116,14 @@ def add_arm(self, arm: Arm, binarizer: Callable = None) -> NoReturn:
         """
         self.arm_to_expectation[arm] = 0
         self._uptake_new_arm(arm, binarizer)
+        self.arm_to_status[arm] = {STATUS_TRAINED: False, STATUS_WARM: False, STATUS_WARM_STARTED_BY: None}
```
skadio, line 119: why not call these IS_TRAINED, IS_WARM, WARM_STARTED_BY? 
May be I am missing sth

mabwiser/base_mab.py
```diff
@@ -394,3 +400,31 @@ def _get_cold_arm_to_warm_arm(self, cold_arm_to_warm_arm, arm_to_features, dista
...
 
         return new_cold_arm_to_warm_arm
+
+    def _reset_arm_to_status(self):
+        self.arm_to_status: Dict[Arm, dict] = {arm: {STATUS_TRAINED: False, STATUS_WARM: False,
+                                                     STATUS_WARM_STARTED_BY: None}
+                                               for arm in self.arms}
+
+    def _set_arms_as_trained(self, decisions: Optional[np.ndarray] = None, is_partial: bool = True):
+        """Sets the given arms as trained, where arms are calculated from the ``decisions``.
```
skadio, line 410: In comments, may be add a word on where is this used?

mabwiser/base_mab.py
```diff
@@ -394,3 +400,31 @@ def _get_cold_arm_to_warm_arm(self, cold_arm_to_warm_arm, arm_to_features, dista
...
+            if arm in arms:
+                # All system arms are now trained
+                self.arm_to_status[arm][STATUS_TRAINED] = True
+
+                # If fitting from scratch, arm is no longer warm started
+                if not is_partial:
+                    self.arm_to_status[arm][STATUS_WARM] = False
+                    self.arm_to_status[arm][STATUS_WARM_STARTED_BY] = None
+
+    def _warm_start(self, arm_to_features: Dict[Arm, List[Num]], distance_quantile: float) -> NoReturn:
```
skadio, line 425: So the public warm_start call become abstract now? And this implements _warm_start to be called from LPs? 

Some commentary on the usage of this private _warm_start would help in future

mabwiser/mab.py
```diff
@@ -969,6 +969,18 @@ def neighborhood_policy(self):
         else:
             return None
 
+    @property
+    def cold_arms(self) -> List[Arm]:
+        if not self.neighborhood_policy:
+            # No neighborhood policy, cold arms are calculated at the learning policy level
+            return self._imp.cold_arms
+
+        else:
```
skadio, line 978: Design Q: if not defined, should return empty List() or Abort with some exception? No inclination one way or the other..

mabwiser/neighbors.py
```diff
@@ -75,12 +75,9 @@ def predict_expectations(self, contexts: np.ndarray = None) -> Union[Dict[Arm, N
         return self._parallel_predict(contexts, is_predict=False)
 
     def warm_start(self, arm_to_features: Dict[Arm, List[Num]], distance_quantile: float):
-        # Can only execute warm start when learning policy has been fit in _get_nhood_predictions
```
skadio, line 78: It seems we had "some" definition of what WS is when NP is used? Any idea what it was?

tests/test_mab.py
```diff
@@ -1767,3 +1767,131 @@ def test_pickle_fitted_reproducibile_expectations(self):
                 self.assertListEqual(pe1, pe2)
 
         os.remove('mab.pkl')
+
+    #################################################
+    # Test arm_status
```
skadio, line 1772: Thanks for these tests. Maintaining arm_status correctly is non-trivial in mind. Since a "mab" object is a living object, many things can happen on top if.. later in the process, having arm_status to still be correct requires some good testing at this stage possibly with interesting usage scenarios.

skadio: Just to set expectations; I expect to run into bugs in future where we could not track arm_status correctly.

skadio (dismissed):
LGTM. Minor comments

mabwiser/base_mab.py
bkleyn: When arm is partial fit `is_trained` will change to True, but we keep `is_warm` as True together with info in `warm_started_by` in order to track that the arm was originally warm started.

mabwiser/base_mab.py
bkleyn: I'll update

mabwiser/base_mab.py
bkleyn: I'll update

mabwiser/base_mab.py
bkleyn: I'll add

mabwiser/mab.py
bkleyn: I think we are de-facto saying that NPs do not have any cold arms so returning empty list makes sense to me.

mabwiser/neighbors.py
bkleyn: Yes we made an attempt. Basically, for Neighbors it would warm start the LP that is trained on all the neighbor contexts. Note that this happens during predict so we cannot really say if the MAB has any cold arms or not since it will depend on the context being predicted. We had something similar for Clusters, the implementation details differ.

tests/test_mab.py
bkleyn: Similar concerns. I'll add some more tests.

mabwiser/base_mab.py
bkleyn: Correct. I'll add comments.

mabwiser/base_mab.py
skadio: My bad.. I quickly skimmed and missed that "status" is actually a dictionary itself.. Thought it was singleton. keeping the history makes total sense. May be a quick pydoc comment on exactly what you have above for that warm+fine-tuned case.

tests/test_mab.py
skadio: and I am totally OK with that expectation.. The only issue/worry, if used in PROD, then such bugs will come back to us.. Some calculated risk, I guess..

skadio (approved):
LGTM, and really like the idea of "Resolving Thread" as comments are fixed/taken into account. It cleans up the PR view and makes it easier to come back to the same PR for another pass (because now I only see "remaining" things) Good practice to follow in future and distill to the team

bkleyn:
@skadio thank you for review! I addressed the final few comments and added some additional tests. I'm going to update the docs and merge.

bkleyn (approved):

PR #66, opened 2023-02-06
Vectorized Predict function for Linear Policies 

nagireddyakshay:
### Changes/Optimizations
- Vectorized the  _predict_contexts() function across all Linear Policies and added a new fn  **_vectorized_predict_context()**

### Outcomes
- The revised code is 6x faster than the original implementation

### Reproducibility
- Old fn creates a new random object for each context provided for reproducibility as the fn runs in parallel
- New fn don't require any of the above and utilizes only one random object for all the contexts based of a seed value
- **Outcome**- Reproducability retained

### Changes to the codebase
- The values in the assert fn across LinTS, LinGreedy and Parallel tests were modified to accommodate changes from the new fn
- **Cause** -The use of single seed value across all the contexts produces a different reproducible result


mabwiser/linear.py
```diff
@@ -90,50 +90,47 @@ def predict(self, x):
...
-    def predict(self, x):
-
+    def predict(self, x: np.ndarray):
         # Scale
         if self.scaler is not None:
             x = self._scale_predict_context(x)
 
         # Randomly sample coefficients from multivariate normal distribution
         # Covariance is enhanced with the exploration factor
-        beta_sampled = self.rng.multivariate_normal(self.beta, np.square(self.alpha) * self.A_inv)
```
skadio, line 114: what happened here?

mabwiser/linear.py
```diff
@@ -90,50 +90,47 @@ def predict(self, x):
...
+        beta_sampled = self.rng.multivariate_normal(self.beta, np.square(self.alpha) * self.A_inv, size=x.shape[0])
 
         # Calculate expectation y = x * beta_sampled
-        return np.dot(x, beta_sampled)
+        return np.sum(x * beta_sampled, axis=1)
```
skadio, line 113: is this the same, but faster?

mabwiser/linear.py
```diff
@@ -90,50 +90,47 @@ def predict(self, x):
...
         # Scale
         if self.scaler is not None:
             x = self._scale_predict_context(x)
 
+        # Calculating x_A_inv
+        x_A_inv = np.dot(x, self.A_inv)
+
         # Upper confidence bound = alpha * sqrt(x A^-1 xt). Notice that, x = xt
-        ucb = (self.alpha * np.sqrt(np.dot(np.dot(x, self.A_inv), x)))
+        ucb = self.alpha * np.sqrt(np.sum(x_A_inv * x, axis=1))
```
skadio, line 127: faster?

mabwiser/linear.py
```diff
@@ -208,40 +205,38 @@ def _fit_arm(self, arm: Arm, decisions: np.ndarray, rewards: np.ndarray, context
...
-            rng = create_rng(seed=seeds[index])
+        # Initializing array with expectations for each arm
+        num_contexts = contexts.shape[0]
+        arm_expectations = np.empty((num_contexts, len(arms)), dtype=float)
 
-            # With epsilon probability set arm expectation to random value
-            if rng.rand() < self.epsilon:
-                for arm in arms:
-                    arm_to_expectation[arm] = rng.rand()
+        # With epsilon probability, assign random flag to context
```
skadio, line 220: does this mean behavior of existing test/code will change? Since we are changing from 1by1 random setc?

skadio (approved):
LGTM!

mabwiser/linear.py
nagireddyakshay: Nope. This change was done to accomodate multiple inputs/contexts at once and calulate the ucb value respectively. Now both the input- x and -the variable ucb are of np.ndarray type

mabwiser/linear.py
nagireddyakshay: This change was done to accomodate multiple inputs/contexts at once and calulate the expectation value respectively. Now both the input- x and the function return type are of type np.ndarray

mabwiser/linear.py
nagireddyakshay: No, the behaviour would be the same from the algorithmic point of view. The reason we were using a random state for each context was to factor in the seed values for multi-processing module for results reproducibility. Since, we have eliminated the need for multi-processing using NumPy vectorization, we can just use one global seed value for reproducibility.

mabwiser/linear.py
nagireddyakshay: The above line of code draws random samples from a multivariate normal distribution for each context individually. So this line was modified to generate the random samples for all the contexts in one single go.

mabwiser/linear.py
skadio: Ah I see! thanks

mabwiser/linear.py
skadio: These are good commentary (ex; generate random samples for all contexts in one single go) ee if you can add a bit of this as code comments, if not already

mabwiser/linear.py
skadio: Gotch, by _behavior_ I meant the output and specific cases. I see that tests are tweaked a little with the new behavior from one global seed. Cool!

bkleyn (approved):

PR #82, opened 2023-08-02
Fixes #75

### LINKED ISSUE #75: Need an LP and NP Type Definition
In the constructor of MAB, we use a Union of LPs and Union of NPs written explicitly. 

https://github.com/fidelity/mabwiser/blob/master/mabwiser/mab.py#L732

This is problematic.. 

This creates an issue for ALSN not being able to a generic LP Type or NP Type, unless it also explicity states all options as a Union -- which is not good. 

This issue is to create a Type Defintion, LPType and NPType similar to ARM type  https://github.com/fidelity/mabwiser/blob/master/mabwiser/utils.py#L13

This can simplify the MAB() constuctor using these type. And... ALNS and other libraries on top, can refer to these types without necessarily knowing what is in it. When mabwiser introduces new policies, pip upgrading the library at the top-level library will uptake new types automatically. 

Note: once this features is complete, it also requires a minor code change in the ALNS to refer to these new types in the MABSelector()

AshishPvjs:
Code for fixing the Issue #75

### DIFF
--- a/mabwiser/mab.py
+++ b/mabwiser/mab.py
@@ -750,7 +750,7 @@ class MAB:
 
     def __init__(self,
                  arms: List[Arm],  # The list of arms
-                 learning_policy: LearningPolicyTypeNew,  # The learning policy
+                 learning_policy: LearningPolicyType,  # The learning policy
                  neighborhood_policy: NeighborhoodPolicyType = None,  # The context policy, optional
                  seed: int = Constants.default_seed,  # The random seed
                  n_jobs: int = 1,  # Number of parallel jobs
--- a/mabwiser/mab.py
+++ b/mabwiser/mab.py
@@ -10,7 +10,7 @@ This module defines the public interface of the **MABWiser Library** providing a
     - ``NeighborhoodPolicy``
 """
 
-from typing import List, Union, Dict, NamedTuple, NoReturn, Callable, Optional
+from typing import List, Union, Dict, NamedTuple, NoReturn, Callable, Optional, NewType
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


mabwiser/mab.py
```diff
@@ -667,6 +667,25 @@ def _is_compatible(self, learning_policy: LearningPolicy):
                                                 LearningPolicy.ThompsonSampling))
 
 
+LearningPolicyType = NewType('LearningPolicyType', Union[LearningPolicy.EpsilonGreedy,
```
bkleyn, line 670: Mind adding a brief comment like "LearningPolicyType is the union of all learning policies"

mabwiser/mab.py
```diff
@@ -667,6 +667,25 @@ def _is_compatible(self, learning_policy: LearningPolicy):
...
+                                                         LearningPolicy.Random,
+                                                         LearningPolicy.Softmax,
+                                                         LearningPolicy.ThompsonSampling,
+                                                         LearningPolicy.UCB1,
+                                                         LearningPolicy.LinGreedy,
+                                                         LearningPolicy.LinTS,
+                                                         LearningPolicy.LinUCB])
+
+
+NeighborhoodPolicyType = NewType('NeighborhoodPolicyType', Union[None,
```
bkleyn, line 681: Also add comment for NeighborhoodPolicyType

bkleyn:
Note code in `tests/test_base.py` lines 78 and 82 that might be possible to replace with the new types.

skadio (approved):
Looks great, and I assume all tests pass?

skadio (approved):

skadio (dismissed):

skadio (approved):

PR #91, opened 2024-02-16
Remove incorrect use of `NoReturn`

LiHRaM:
Hi, great library! While developing, I noticed that many of the functions are annotated with `NoReturn`, which means static analyzers are expected to mark any code after calling those functions as unreachable.

See [PEP 0484](https://peps.python.org/pep-0484/#the-noreturn-type) for documentation on uses of `NoReturn`.

---

**TL;DR**: this library uses `NoReturn` in places where you would expect `None` to be the return type, because the code is expected to continue after execution of functions, as illustrated in the examples.

---

Wasn't sure about updating the documentation, but decided to deliver the whole thing in one PR.
I updated the docs by running `make github` in the `docsrc/` directory after installing `requirements.txt`, but I'm not sure if that was the process meant to be used or if it was supposed to be something else.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,6 +2,11 @@
 MABWiser CHANGELOG
 =====================
 
+February, 05, 2024 2.7.2
+-------------------------------------------------------------------------------
+minor:
+- Fixed default KMeans n_init parameters instead of using 'auto' used in scikit-learn>=1.4
+
 August, 02, 2023 2.7.1
 -------------------------------------------------------------------------------
 minor:
--- a/mabwiser/_version.py
+++ b/mabwiser/_version.py
@@ -3,5 +3,5 @@
 
 __author__ = "FMR LLC"
 __email__ = "opensource@fmr.com"
-__version__ = "2.7.1"
+__version__ = "2.7.2"
 __copyright__ = "Copyright (C), FMR LLC"
--- a/mabwiser/clusters.py
+++ b/mabwiser/clusters.py
@@ -27,9 +27,9 @@ class _Clusters(BaseMAB):
         self.n_clusters = n_clusters
 
         if is_minibatch:
-            self.kmeans = MiniBatchKMeans(n_clusters, random_state=rng.seed)
+            self.kmeans = MiniBatchKMeans(n_clusters, random_state=rng.seed, n_init=3)
         else:
-            self.kmeans = KMeans(n_clusters, random_state=rng.seed)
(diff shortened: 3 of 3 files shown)


docs/_sources/new_bandit.rst.txt
```diff
@@ -113,7 +113,7 @@ Here is, at a high-level, what you need to implement in your bandit policy:
             # These fields are declared here and initialized to zero.
             self.my_value_to_arm = dict.fromkeys(self.arms, 0)
 
-        def fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray = None) -> NoReturn:
+        def fit(self, decisions: np.ndarray, rewards: np.ndarray, contexts: np.ndarray = None):
```
skadio, line 116: May I suggest that, we return a "None" instead of the incorrect NoReturn 

from () -> NoReturn to () -> None

This is clearly a miss on our part, so we highly appreciate your fix!

docs/_static/_sphinx_javascript_frameworks_compat.js
```diff
@@ -0,0 +1,123 @@
+/* Compatability shim for jQuery and underscores.js.
```
skadio, line 1: what's this completely new file? is this needed?

mabwiser/approximate.py
```diff
@@ -5,7 +5,7 @@
 from collections import defaultdict
 from copy import deepcopy
 from itertools import chain
-from typing import List, NoReturn, Optional, Union
+from typing import List, Optional, Union
```
skadio, line 8: We can import "None" instead of NoReturn

skadio (changes requested):
@LiHRaM this is great, thank you so much for the fix! 

This is clearly a miss on our part and your fix will help us and everyone. 

Only minor request/comments: 
- can we please return None instead to replace the NoReturn? 
- can we please add comment line in the CHANGELOG.txt with a minor update with sth like "replaced NoReturn with None"? Please also include a sentence on "contributed by @LiHRaM" so that we can acknowledge your work. 
- Along that line, please upgrade the _version.py to 2.7.3 please
- I noticed a net-new doc file added, is this needed? 
- Could you confirm that all tests still pass?

docs/_static/_sphinx_javascript_frameworks_compat.js
LiHRaM: I think the  difference in `_static` is because the versions of sphinx and the theme used were not pinned, and so newer versions have newer dependencies etc.

